PR #1, opened 2017-12-06
DOC: Add more details on the dev plan to the README

shoyer:


README.md
```diff
@@ -1,2 +1,22 @@
-# numpy_stubs
-Experimental typing stubs for NumPy
+# numpy_stubs: experimental typing stubs for NumPy
+
+This repository exists for developing [PEP 484](https://www.python.org/dev/peps/pep-0484/)
+compatible typing annotations for NumPy.
```
eric-wieser: Might be nice to link `NumPy`, even though everyone knows what it is

shoyer:
> Presumably, the eventual plan, once it stabilizes, and numpy drops python 2, would for this to be part of the numpy source code itself?

Yes, though I don't think that's contingent on Python 3. Type stubs in separate files can exist even under Python 2.

PR #4, opened 2017-12-06
Add basic scaffolding for distributing the package

### LINKED ISSUE #2: Setup basic scaffolding (e.g., setup.py) for distributing this package

eric-wieser:
It doesn't appear that the released mypy supports _stub packages right now, so for now we have to add to MYPYPATH manually.

Something else to consider is how to sync versions with numpy, but we can deal with that later, if we start our numbers low enough.

This adds a very small number of stubs just to verify that installation maybe works.

Fixes #2.

### DIFF
--- a/numpy/__init__.pyi
+++ b/numpy/__init__.pyi
@@ -1,0 +1,12 @@
+# very simple, just enough to start running tests
+
+class ndarray: pass
+
+class dtype: pass
+
+def array(
+    object: object,
+    dtype: dtype = ...,
+    copy: bool = ...,
+    subok: bool = ...,
+    ndmin: int = ...) -> ndarray: ...
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,18 @@
+from setuptools import setup, find_packages
+
+setup(
+    name='numpy_stubs',
+    maintainer="NumPy Developers",
+    maintainer_email="numpy-discussion@python.org",
+    description="PEP 561 type stubs for numpy",
+    url="http://www.numpy.org",
+    license='BSD',
+    version="0.0.1",
+    packages=find_packages(),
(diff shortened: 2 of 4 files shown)


numpy/__init__.pyi
```diff
@@ -0,0 +1,12 @@
+# very simple, just enough to start running tests
+
+class ndarray: pass
+
+class dtype: pass
+
+def array(
+    object : object,
```
shoyer: nit: PEP8 suggests not using a space before `:`

numpy/__init__.pyi
```diff
@@ -0,0 +1,12 @@
+# very simple, just enough to start running tests
+
```
shoyer: We should add annotations `def __getattr__(name) -> Any: ...` to the module and `def __getattr__(self, name) -> Any: ...` to `ndarray`/`dtype` to mark them as flexibly typed.

numpy/__init__.pyi
eric-wieser: Oh, it does? Shoot, I've been doing that wrong everywhere in my own code.

shoyer: See "Function annotations" under http://legacy.python.org/dev/peps/pep-0008/#other-recommendations

numpy/__init__.pyi
eric-wieser: Not sure what you mean about adding it to the module. Is that supported in type stubs? It's certain not supported in normal python,

shoyer: Well, OK, we can leave it out for now. It is indeed valid in stub files (only).

eric-wieser:
Updated to address PEP8

Issue #3, opened 2017-12-06
Continuous integration tests


shoyer:
What should these look like?

At the least, we probably should verify that mypy can successfully type-check some simple examples using these type annotations.

henryJack:
Happy to write the example annotated code once the stubs are done. It's probably best to add a mypy job to the pipeline with multiple scripts testing that mypy passes and throws errors where it's supposed to.

eric-wieser:
I'm workin on #2, and I'm not even sure I'm able to test this locally - it seems `mypy` doesn't support `_stubs` packages yet.

shoyer:
It seems that we are very cutting edge here in relying on PEP 561. @ethanhs is still finishing up the mypy implementation in https://github.com/python/mypy/pull/4278, but I guess we could test locally against his branch.

shoyer:
See https://github.com/python/typeshed/issues/754 and https://github.com/python/typeshed/issues/1339. It seems like tests for stubs make sense for at least tricky cases, but are probably overkill when required for everything.

eric-wieser:
We're probably going to want to use whatever mypy is using to run its own tests here. Perhaps to support PEP561, mypy should extract its testing module to be standalone

ethanhs:
The branch is in a somewhat limited state, it checks against packages installed only for the Python mypy is installed in. Just a heads up. It should work for the simple case.

shoyer:
We have this now.

PR #8, opened 2018-02-18
Setup Travis and correct setup.py

alanhdu:
This doesn't really take an opinion on #3, but it at least sets up the basic sanity checks that this repo already has.


setup.py
```diff
@@ -13,6 +13,6 @@
     # PEP 561 requires these
     install_requires=['numpy~=1.13.0'],
```
shoyer: shall we update this to 1.14.0?

shoyer:
Can you also add a Travis-CI link to the README?

    [Build Status](https://travis-ci.org/numpy/numpy_stubs)

setup.py
```diff
@@ -1,7 +1,7 @@
 from setuptools import setup, find_packages
 
 setup(
-    name='numpy_stubs',
+    name='numpy-stubs',
```
shoyer: I see that PEP 561 changed it's guidelines here. I suppose we should also rename this repository?

ethanhs: This name (the distribution name) doesn't impact PEP 561 compliance. The *package name* is the real decider; when this repo starts using my PEP 561 implementation in mypy, it will need to rename the `numpy` package to `numpy-stubs`, and make according changes. Happy to do those changes when it is time.

PR #9, opened 2018-02-18
Add basic stubs for some class attributes

alanhdu:


numpy/__init__.pyi
```diff
@@ -1,8 +1,91 @@
...
+    names: Optional[Tuple[str, ...]]
+    num: int
+    shape: Tuple[int, ...]
+    ndim: int
+    subdtype: Optional[Tuple[dtype, Tuple[int, ...]]]
+
+
+    def newbyteoder(self, new_order:str = ...): dtype
+
+    str: builtins.str
```
alanhdu: I've segregated this here so that the other `str`s don't need to have `builtins.str` here (see https://github.com/python/mypy/issues/3775).

numpy/__init__.pyi
```diff
@@ -1,8 +1,91 @@
...
+from typing import Any, Mapping, List, Optional, Tuple, Union
 
-class ndarray: pass
+from numpy.core._internal import _ctypes
+
+class dtype:
+    alignment: int
+    byteorder: str
+    char: str
+    descr: List[Tuple[str, str]]
```
shoyer: Entries in `descr` can only be three element tuples, with the last item indicating a shape, e.g., 
```
>>> np.dtype([('x', 'f4'), ('y', np.float32), ('z', 'f4', (2,2))]).descr
[('x', '<f4'), ('y', '<f4'), ('z', '<f4', (2, 2))]
```

numpy/__init__.pyi
```diff
@@ -1,8 +1,91 @@
...
+    fields: Optional[Mapping[str, Union[Tuple[dtype, int], Tuple[dtype, int, str]]]]
```
shoyer: It looks like the final `title` field can have any type, so it should probably be keyed as `Any` instead of `str`: https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.dtype.fields.html#numpy.dtype.fields

numpy/__init__.pyi
shoyer: Please add this as a comment in the code

numpy/__init__.pyi
```diff
@@ -1,8 +1,91 @@
...
+    kind: str
+    name: str
+    names: Optional[Tuple[str, ...]]
+    num: int
+    shape: Tuple[int, ...]
+    ndim: int
+    subdtype: Optional[Tuple[dtype, Tuple[int, ...]]]
+
+
+    def newbyteoder(self, new_order:str = ...): dtype
```
shoyer: please keep the space after `:` per pep8. This should be `new_order: str = ...`

shoyer: please keep the space after `:` per pep8. This should be:

    new_order: str = ...

numpy/__init__.pyi
```diff
@@ -1,8 +1,91 @@
...
+    isalignedstruct: bool
+    itemsize: int
+    kind: str
+    name: str
+    names: Optional[Tuple[str, ...]]
+    num: int
+    shape: Tuple[int, ...]
+    ndim: int
+    subdtype: Optional[Tuple[dtype, Tuple[int, ...]]]
+
```
shoyer: It isn't documented, but `metadata` is also used out in the wild (e.g., by h5py). It should be `Optional[Mapping]`.

numpy/__init__.pyi
```diff
@@ -1,8 +1,91 @@
...
+
+    def newbyteoder(self, new_order:str = ...): dtype
+
+    str: builtins.str
+    type: builtins.type
+
+    @property
+    def base(self) -> dtype: ...
```
shoyer: Is there a reason why this is a property vs. an attribute for everything else? Does mypy care about the difference?

numpy/__init__.pyi
```diff
@@ -1,8 +1,91 @@
...
+
+
+_dtype = dtype      # for ndarray type
+
+
+class flagsobj:
```
shoyer: `flagsobj` is not exposed by NumPy's public API, so let's give it a private name (`_flagsobj`?)

numpy/__init__.pyi
```diff
@@ -1,8 +1,91 @@
...
+    def newbyteoder(self, new_order:str = ...): dtype
+
+    str: builtins.str
+    type: builtins.type
+
+    @property
+    def base(self) -> dtype: ...
+
+
+_dtype = dtype      # for ndarray type
```
shoyer: Give this a name like `_dtype_class` to make it a little clearer what it is?

numpy/__init__.pyi
```diff
@@ -1,8 +1,91 @@
...
+    def __iter__(self) -> flatiter: ...
+    def __next__(self) -> Any: ...
+
+
+class ndarray:
+    T: ndarray
+    data: memoryview
+    dtype: _dtype
+    flags: flagsobj
+    flat: Any
```
shoyer: This is a `flatiter` object

numpy/__init__.pyi
```diff
@@ -1,8 +1,91 @@
...
+    imag: ndarray
+    real: ndarray
+    size: int
+    itemsize: int
+    nbytes: int
+    ndim: int
+    shape: Tuple[int, ...]
+    strides: Tuple[int, ...]
+    ctypes: _ctypes
+    base: Optional[ndarray]
```
shoyer: This list is incomplete for now, so let's add `def __getattr__(self, name) -> Any: ...`.

numpy/__init__.pyi
JelleZijlstra: I assume this should be `newbyteorder`, not `newbyteoder`?

numpy/__init__.pyi
JelleZijlstra: Properties can't be assigned to unless they have a setter. With this stub, if you had `x: dtype`, mypy would allow `x.type = some_type` but not `x.base = some_dtype`.

shoyer: OK, in that case we should also convert most of the other attributes into readonly properties.

alanhdu:
This should be ready for another round of review!

numpy/__init__.pyi
```diff
@@ -1,8 +1,186 @@
...
+    @property
+    def byteorder(self) -> str: ...
+
+    @property
+    def char(self) -> str: ...
+
+    @property
+    def descr(self) -> List[Union[
+        Tuple[str, str],
+        Tuple[str, str, Tuple[int, ...]]]]: ...
```
shoyer: optional: it might be convenient to define an alias `_Shape = Tuple[int, ...]`

numpy/__init__.pyi
```diff
@@ -1,8 +1,186 @@
...
+from typing import Any, Mapping, List, Optional, Tuple, Union
 
-class ndarray: pass
+from numpy.core._internal import _ctypes
+
+class dtype:
+    names: Optional[Tuple[str, ...]]
+    updateifcopy: bool
+    writeable: bool
+    writebackifcopy: bool
```
shoyer: I'm not quite sure where these attributes came from -- I can't find any of `updateifcopy`, `writeable` and `writebackifcopy` on dtype.

shoyer: Actually, these are missing down below on flagsobj.

numpy/__init__.pyi
```diff
@@ -1,8 +1,186 @@
...
+
+    def newbyteorder(self, new_order: str = ...) -> dtype: ...
+
+    # Leave str and type for end to avoid having to use `builtins.str`
+    # everywhere. See https://github.com/python/mypy/issues/3775
+    @property
+    def str(self) -> builtins.str: ...
+
+    @property
+    def type(self) -> builtins.type: ...
```
shoyer: More specifically, we  could define this as the `np.generic` class update and  switch this to `Type[generic]`.

numpy/__init__.pyi
alanhdu: Oops :blush:

shoyer:
Thanks @alanhdu !

numpy/__init__.pyi
```diff
@@ -1,8 +1,188 @@
...
+
+
+_dtype_class = dtype      # for ndarray type
```
eric-wieser: Why does this have a duplicate name?

shoyer: Otherwise mypy gets confused when you to type the `dtype` ndarray attribute, e.g., `def dtype(self) -> dtype` refers to `ndarray.dtype` for the return value, not the `dtype` class.

eric-wieser: Can you fix it with an `import numpy as np`, and then using `np.dtype`?

shoyer: I don't know exactly how mypy handles circular imports, but I suspect that importing a module from itself would not work.

PR #10, opened 2018-02-20
Define special methods for ndarray and add more extensive tests.

shoyer:


tests/test_simple.py
```diff
@@ -1,5 +1,48 @@
...
+len(array)
+str(array)
+array + 1
+-array
 
-def foo(a: np.ndarray): pass
+array_scalar = np.array(1)
+int(array_scalar)
+float(array_scalar)
+complex(array_scalar)
```
shoyer: mypy current errors on this line: `tests/test_simple.py:41: error: No overload variant of "complex" matches argument types [numpy.ndarray]`

Is this a mypy bug or am I doing something wrong?

ethanhs: I believe this is https://github.com/python/mypy/issues/2136, in other words, you need to add `SupportsComplex` as a base since mypy doesn't yet infer protocols based on dunder methods.

shoyer: Reported at https://github.com/python/mypy/issues/4592

shoyer:
@eric-wieser Want to take a look?

shoyer:
@alanhdu any thoughts?

numpy/__init__.pyi
```diff
@@ -181,12 +184,114 @@ class ndarray:
...
+    def __getitem__(self, key) -> Any: ...
+    def __setitem__(self, key, value): ...
+    def __iter__(self) -> Any: ...
+    def __contains__(self, key) -> bool: ...
+
+    def __int__(self) -> int: ...
+    def __float__(self) -> float: ...
+    def __complex__(self) -> complex: ...
+    def __oct__(self) -> str: ...
+    def __hex__(self) -> str: ...
```
eric-wieser: These should be conditional on the python version. I think mypy supports that?

tests/test_simple.py
```diff
@@ -1,5 +1,49 @@
...
+array_2d = np.ones((3, 3))
+array_2d[:2, :2]
+array_2d[..., 0]
+array_2d[:2, :2] = 0
+
+# Other special methods
+len(array)
+str(array)
+array + 1
+-array
```
eric-wieser: Do you want to test all the binary operators here too?

numpy/__init__.pyi
```diff
@@ -181,12 +184,114 @@ class ndarray:
...
+    def __invert__(self) -> ndarray: ...
+
+    # TODO(shoyer): remove when all methods are defined
     def __getattr__(self, name) -> Any: ...
 
 
 def array(
     object: object,
-    dtype: dtype = ...,
+    dtype: _dtype_class = ...,
```
eric-wieser: `_dtype_class` is "things which can be converted to a dtype"?

numpy/__init__.pyi
ethanhs: Yes, it supports limited version checking with `sys.version_info`.

numpy/__init__.pyi
shoyer: No, it's literally the `dtype` class. This should be relaxed to something more flexible, e.g., `Union[Type[np.generic], str, dtype]` (plus probably some other types for structured dtypes).

tests/test_simple.py
shoyer: Yeah, that's probably a good idea.

numpy/__init__.pyi
```diff
@@ -196,8 +222,9 @@ class ndarray(Iterable, Sized, SupportsInt, SupportsFloat, SupportsComplex,
     def __int__(self) -> int: ...
     def __float__(self) -> float: ...
     def __complex__(self) -> complex: ...
-    def __oct__(self) -> str: ...
-    def __hex__(self) -> str: ...
+    if sys.version_info.major < 3:
+        def __oct__(self) -> str: ...
+        def __hex__(self) -> str: ...
     def __nonzero__(self) -> bool: ...
```
eric-wieser: This is called __bool__ on py3

numpy/__init__.pyi
```diff
@@ -83,7 +109,8 @@ class dtype:
     def type(self) -> builtins.type: ...
 
 
-_dtype_class = dtype      # for ndarray type
+_DtypeLike = Union[dtype, _ConvertibleToDtype]
+_Dtype = dtype  # to avoid name conflicts with ndarray.dtype
```
eric-wieser: It would be great if we could have a "strict-typing" mode that narrowed the scope of what dtype-like or array-like means. I wonder how that could be achieved...

numpy/__init__.pyi
```diff
@@ -1,15 +1,41 @@
...
 from numpy.core._internal import _ctypes
 
 _Shape = Tuple[int, ...]
 
+# Anything that can be coerced into numpy.dtype. To avoid recursive
+# definitions, any nested fields are required to be castable to a dtype object
+# are typed as Any.
+# Refernce: https://docs.scipy.org/doc/numpy/reference/arrays.dtypes.html
+_ConvertibleToDtype = Union[
+    type,  # TODO: enumerate np.generic types and Python scalars
```
eric-wieser: Does `Type[np.generic]` work here?

shoyer: It would, but we haven't defined the scalar type hierarchy yet.

numpy/__init__.pyi
shoyer: Yes, I'd love see this.

numpy/__init__.pyi
```diff
@@ -1,15 +1,42 @@
...
 
 from numpy.core._internal import _ctypes
 
 _Shape = Tuple[int, ...]
 
+# Anything that can be coerced into numpy.dtype. To avoid recursive
+# definitions, any nested fields are required to be castable to a dtype object
+# are typed as Any.
+# Reference: https://docs.scipy.org/doc/numpy/reference/arrays.dtypes.html
+_ConvertibleToDtype = Union[
```
shoyer: Apparently I can actually include a forward reference to `dtype` in this top-level type alias. If that's a standard feature for pyi files, then I suppose we should make use of it rather than defining the separate `_DtypeLike` alias below.

shoyer: Nevermind, this doesn't work.

shoyer: Just pushed a commit implementing this.

numpy/__init__.pyi
```diff
@@ -1,15 +1,43 @@
...
+    type,  # TODO: enumerate np.generic types and Python scalars
+    # TODO: add a protocol for anything with a dtype attribute?
+    None,
+    str,
+    Tuple[Any, int],
+    Tuple[Any, _Shape],
+    List[Union[Tuple[Union[str, Tuple[str, str]], Any],
+               Tuple[Union[str, Tuple[str, str]], Any, _Shape]]],
+    Dict[str, Any],
+    Tuple[Any, Any]]
```
eric-wieser: I think each of these could do with a comment giving an example of the case they aim to cover

numpy/__init__.pyi
```diff
@@ -1,15 +1,52 @@
...
+    None,
+    # array-scalar types and generic types
+    type,  # TODO: enumerate these when we add type hints for numpy scalars
+    # TODO: add a protocol for anything with a dtype attribute
+    # character codes, type strings or comma-separated fields, e.g., 'float64'
+    str,
+    # (flexible_dtype, itemsize)
+    Tuple[Any, int],
+    # (fixed_dtype, shape)
+    Tuple[Any, _Shape],
```
eric-wieser: Note that `(int, 2)` is shorthand for `(int, (2,))`, so this the `int` case covers more than you might expect

numpy/__init__.pyi
```diff
@@ -1,15 +1,52 @@
...
+)
+
+import sys
 
 from numpy.core._internal import _ctypes
 
 _Shape = Tuple[int, ...]
 
+# Anything that can be coerced into numpy.dtype. To avoid recursive
+# definitions, any nested fields are required to be castable to a dtype object
```
eric-wieser: Perhaps do `_DtypeLike_nested = Any  # TODO: wait for support for recursive types`, and then use it in place of `Any` below?

numpy/__init__.pyi
shoyer: OK, in that case I'm going to use `_ShapeLike = Union[int, Tuple[int, ...]]` for clarity.

numpy/__init__.pyi
```diff
@@ -181,12 +225,122 @@ class ndarray:
...
+    def __int__(self) -> int: ...
+    def __float__(self) -> float: ...
+    def __complex__(self) -> complex: ...
+    if sys.version_info[0] < 3:
+        def __oct__(self) -> str: ...
+        def __hex__(self) -> str: ...
+        def __nonzero__(self) -> bool: ...
+    else:
+        def __bool__(self) -> bool: ...
+    def __bytes__(self) -> bytes: ...
```
eric-wieser: I imagine this is python 3 only?

Also `__unicode__`, `__str__`, and `__repr__` are missing

numpy/__init__.pyi
```diff
@@ -1,15 +1,57 @@
...
+    # (fixed_dtype, shape)
+    Tuple[_DtypeLikeNested, _ShapeLike],
+    # [(field_name, field_dtype, field_shape), ...]
+    List[Union[
+        Tuple[Union[str, Tuple[str, str]], _DtypeLikeNested],
+        Tuple[Union[str, Tuple[str, str]], _DtypeLikeNested, _ShapeLike]]],
+    # {'names': ..., 'formats': ..., 'offsets': ..., 'titles': ...,
+    #  'itemsize': ...} or {'field1': ..., 'field2': ..., ...}
+    # TODO: use TypedDict when/if it's officially supported
+    Dict[str, Any],
```
alanhdu: Is this really `Any`? I think there could be:

```python
Union[
    Dict[str, Sequence[Union[int, str, _DtypeLike]]]   # for {"names": ..., "formats": ...}
    Dict[str, Union[Tuple[str, int], Tuple[_DtypeLike, int]]] # for {"field": ....}
]
```

eric-wieser: I think that should be
```

    # for {"names": ..., "formats": ...}
    Dict[str, Union[
        Sequence[str],  #names
        Sequence[_DtypeLike], # formats
        Sequence[int], # offsets
        int,  # itemsize
    ]],
    Dict[str, Tuple[_DtypeLike, int]] # for {"field": ....}
```

JelleZijlstra: This is probably dangerous because dict is an invariant type (so, for example, `Dict[str, int]` wouldn't be compatible with this big union type). Can you use `Mapping` instead?

shoyer: We could use `Mapping`, but it would be inaccurate. NumPy does actually insist on `dict`:
```
In [13]: od = collections.OrderedDict(x=(float, 1), y=(int, 2))

In [14]: np.dtype(od)
Out[14]: dtype({'names':['x','y'], 'formats':['<f8','<i8'], 'offsets':[1,2], 'itemsize':10})

In [15]: np.dtype(ChainMap([od]))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-15-03baf5fab980> in <module>()
----> 1 np.dtype(ChainMap([od]))

TypeError: data type not understood
```

alanhdu: We have the same problem with:

```python
   List[Union[
       Tuple[Union[str, Tuple[str, str]], _DtypeLikeNested],
       Tuple[Union[str, Tuple[str, str]], _DtypeLikeNested, _ShapeLike]]],
```

which maybe should be `Sequence` but NumPy insists on taking a `list` (or at least, it doesn't take a tuple which is my go-to example of a non-list sequence).

numpy/__init__.pyi
shoyer: done

numpy/__init__.pyi
shoyer: I added the more specific types for Dict, which should at least minimize the invariance issue.

Is it possible to make a dict that is covariant in its argument type? e.g.,
```python
K = TypeVar('V', covariant=True)
V = TypeVar('V', covariant=True)
def f(x: Dict[K, V]): ...
```

eric-wieser:
(re the text/Unicode commit)

shoyer:
Flake8 didn't like using `unicode` on Python 3. I guess I could have used a comment to disable the flake8 check but this was an easy work around.

numpy/__init__.pyi
```diff
@@ -181,12 +234,125 @@ class ndarray:
...
+        def __matmul__(self, other): ...
+        def __rmatmul__(self, other): ...
+
+    def __neg__(self) -> ndarray: ...
+    def __pos__(self) -> ndarray: ...
+    def __abs__(self) -> ndarray: ...
+    def __invert__(self) -> ndarray: ...
+
+    # TODO(shoyer): remove when all methods are defined
     def __getattr__(self, name) -> Any: ...
```
eric-wieser: Missing `__setattr__` for `.dtype` and `.shape`?

shoyer: dtype and shape are only typed as attributes, not properties, which means they can be set. But perhaps it would indeed be good to overload setters appropriately...

shoyer:
@eric-wieser @alanhdu thanks for the careful reviews!

PR #11, opened 2018-03-01
[WIP] Typehint np.random

alanhdu:


alanhdu:
This isn't ready yet, but I wanted to get feedback about whether this was the right approach (lots of `@overload`s for type-hinting). The problems I have with it are:

- It's verbose and irritating (I actually have a `jinja2` template that generates this file), although this is a one-time cost so I'm not too concerned.
 - `overload` semantics are a little in flux -- because of https://github.com/python/mypy/issues/4020, some of the overloads are counted as overlapping from MyPy and return an error (see that last commit)
- The type signatures aren't strictly correct -- if you pass in a scalar `ndarray` into one of these functions, you get a scalar non-`ndarray` out. I'm not sure how to deal with this until we can distinguish scalar `ndarray`s and non-scalar `ndarray`s.

I'm not sure what the alternative approaches are -- I thought about trying to create a special `_Distribution`  that subclasses `Callable` and somehow does all of the broadcasting for us, but I couldn't actually figure out how to get that to work.

numpy/random/mtrand.pyi
```diff
@@ -0,0 +1,132 @@
...
+               ) -> _T: ...
+    @overload
+    def choice(self, a: Union[int, Sequence[_T]], size: _Size,
+               replace: bool = ..., p: Optional[Sequence[float]] = ...
+               ) -> ndarray: ...
+
+    @overload
+    def beta(self, a: float, b: float) -> float: ...
+    @overload
+    def beta(self, a: float, b: float, size: _Size) -> float: ...
```
shoyer: If `size`  is specified, you get an `ndarray`, not a float.

numpy/random/mtrand.pyi
```diff
@@ -0,0 +1,132 @@
...
+    @overload
+    def beta(self, a: ndarray, b: float) -> ndarray: ...
```
shoyer: Unfortunately this isn't quite right -- if you pass in a zero-dimensional ndarray, you get a float back, not a ndarray:
```
In [17]: rs.beta(np.array(2), 2)
Out[17]: 0.3340049666412606
```

This sort of behavior (zero dimensional arrays -> scalars) is unfortunately quite common with NumPy.

numpy/random/mtrand.pyi
```diff
@@ -0,0 +1,132 @@
...
+class RandomState:
+    def __init__(self, state: Optional[Union[int, ndarray]] = ...) -> None: ...
+    def bytes(self, length: int) -> builtins.bytes: ...
+    def dirichlet(self, alpha: Sequence[float], size: Optional[_Size] = ...
+                  ) -> ndarray: ...
+
+    @overload
+    def choice(self, a: int, *,
+               replace: bool = ..., p: Optional[Sequence[float]] = ...
+               ) -> int: ...
```
shoyer: This is `np.int64`, not `int` (at least not on Python 3)

numpy/random/mtrand.pyi
```diff
@@ -0,0 +1,132 @@
...
+    @overload
+    def choice(self, a: Sequence[_T], *,
+               replace: bool = ..., p: Optional[Sequence[float]] = ...
+               ) -> _T: ...
```
shoyer: `a` gets cast to an ndarray, so the return value is the corresponding NumPy scalar type.

numpy/random/mtrand.pyi
```diff
@@ -0,0 +1,132 @@
+import builtins
+from typing import overload, Optional, Sequence, Tuple, TypeVar, Union
+
+from numpy import ndarray
+
+_Size = Union[int, Tuple[int, ...]]
+_T = TypeVar('_T')
+
+class RandomState:
+    def __init__(self, state: Optional[Union[int, ndarray]] = ...) -> None: ...
```
shoyer: The constructor argument is called `seed`, not `state`:
https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.RandomState.html

Also, it can accept sequences of integers ("array_like") not just ndarray instances.

numpy/random/mtrand.pyi
```diff
@@ -0,0 +1,132 @@
...
+    @overload
+    def beta(self, a: float, b: float) -> float: ...
+    @overload
+    def beta(self, a: float, b: float, size: _Size) -> float: ...
+    @overload
+    def beta(self, a: ndarray, b: float) -> ndarray: ...
+    @overload
+    def beta(self, a: float, b: ndarray) -> ndarray: ...
+    @overload
+    def beta(self, a: ndarray, b: ndarray) -> ndarray: ...
```
shoyer: This method can also accept arbitrary "array-like" types which it coerces to ndarrays, e.g., lists of numbers.

shoyer:
> It's verbose and irritating (I actually have a jinja2 template that generates this file), although this is a one-time cost so I'm not too concerned.

There are going to be quite a few functions like these (e.g., all NumPy ufuncs), so it may make sense to incorporate some sort of template generation in the build process for this library.

I guess ufuncs are objects, we so might be able to specify the type-casting rules once (for each number of function arguments) on the base class. But there are still lots of NumPy functions/methods that aren't ufuncs...

> overload semantics are a little in flux -- because of python/mypy#4020, some of the overloads are counted as overlapping from MyPy and return an error (see that last commit)

This is more concerning to me -- it basically blocks anyone from using our annotations with MyPy. I think we need to ensure that mypy can use our annotations without errors.

> The type signatures aren't strictly correct -- if you pass in a scalar ndarray into one of these functions, you get a scalar non-ndarray out. I'm not sure how to deal with this until we can distinguish scalar ndarrays and non-scalar ndarrays.

Yes, this sort of behavior is very unfortunate for us, and unfortunately is also quite prevalent in NumPy. My *inclination* is that our type-stubs should prioritize correctness more than catching all possible errors. Zero-dimensional arrays are somewhat usual to see in NumPy, but they do come up. Without typing support for array dimensionality, this would mean that many of these return values should be unsatisfying `Any` types.

numpy/random/mtrand.pyi
alanhdu: Hm... I'm not sure how to represent "array-like" types. A naive approach would be something like:

```python
_ArrayLike = Union[Sequence[_T], ndarray]
```

But nested sequences are also `_ArrayLike` and MyPy doesn't have support for recursive types (which we'd need for arbitrarily nested sequences).

alanhdu:
I believe I've figured out how to hack around the `@overload` problem -- by explicitly annotating `size: None`, MyPy no longer thinks that they're conflicting types.

> There are going to be quite a few functions like these (e.g., all NumPy ufuncs), so it may make sense to incorporate some sort of template generation in the build process for this library.

Yeah, I think some template generation "build" makes a lot of sense. my approach w/ Jinja2 is pretty ugly (because of how it handles whitespace) and still fails `flake8` despite my best efforts. Do you know of a good way to template out the AST level (or something that's not whitespace sensitive?)? For prettifying, I've thought about using [yapf](https://github.com/google/yapf) or some other autoformatter (although we might have to tweak them for `.pyi` files).

> Yes, this sort of behavior is very unfortunate for us, and unfortunately is also quite prevalent in NumPy. My inclination is that our type-stubs should prioritize correctness more than catching all possible errors. Zero-dimensional arrays are somewhat usual to see in NumPy, but they do come up. Without typing support for array dimensionality, this would mean that many of these return values should be unsatisfying Any types.

I'm willing to defer to your judgement here, but I'd personally lean the other way -- IME w/ MyPy is still limited enough that you almost never have "seamless" integration with any reasonably large codebase, and given that you have to adapt the coding style (to write "type-friendly" code)  and do integration work anyways, I think it's reasonable to prioritize "correctness" (as long as the burden's not too much, like having to add some `int` and `float` casts for numpy scalars).

In any case, I think `Any` is an ok return type -- we could also return a `Union[int, ndarray]` or whatever, although that's arguably the worst of both worlds (since you'd almost certainly need to do a type assertion on the output anyways).

person142:

PR #14, opened 2018-03-20
Add NumPy scalar hierarchy

alanhdu:


numpy/__init__.pyi
```diff
@@ -358,6 +358,187 @@ class ndarray(Iterable, Sized, SupportsInt, SupportsFloat, SupportsComplex,
     def __getattr__(self, name) -> Any: ...
 
 
+_Scalar = TypeVar("_Scalar", bound="generic")
+
+class generic:
+    @property
```
shoyer: Could we define a private base class for all the shared methods/properties between `generic` and `ndarray`? Otherwise we'll have quite a bit to keep in sync.

numpy/__init__.pyi
```diff
@@ -358,6 +358,187 @@ class ndarray(Iterable, Sized, SupportsInt, SupportsFloat, SupportsComplex,
...
+class object_(generic): ...
+class datetime64(generic): ...
+
+class integer(number): ...
+class signedinteger(integer): ...
+class int8(signedinteger): ...
+class int16(signedinteger): ...
+class int32(signedinteger): ...
+class int64(signedinteger): ...
+int_ = int64
```
shoyer: Regrettably, this is actually platform dependent, e.g., on Windows, it's int32: https://github.com/numpy/numpy/issues/9464

numpy/__init__.pyi
```diff
@@ -358,6 +358,187 @@ class ndarray(Iterable, Sized, SupportsInt, SupportsFloat, SupportsComplex,
...
+    def __abs__(self) -> ndarray: ...
+    def __invert__(self) -> ndarray: ...
+
+    # TODO(shoyer): remove when all methods are defined
+    def __getattr__(self, name) -> Any: ...
+
+class number(generic): ...
+class bool_(generic): ...
+class object_(generic): ...
+class datetime64(generic): ...
```
shoyer: Don't forget `timedelta64`. Unfortunately it inherits from `np.integer`, though this is likely to change in the future: https://github.com/numpy/numpy/issues/10685

numpy/__init__.pyi
```diff
@@ -358,6 +358,187 @@ class ndarray(Iterable, Sized, SupportsInt, SupportsFloat, SupportsComplex,
     def __getattr__(self, name) -> Any: ...
 
 
+_Scalar = TypeVar("_Scalar", bound="generic")
```
shoyer: I think we can actually define circular references in `.pyi` files, like I did for `_DtypeLike` above.

numpy/__init__.pyi
```diff
@@ -358,6 +358,187 @@ class ndarray(Iterable, Sized, SupportsInt, SupportsFloat, SupportsComplex,
...
+
+class generic:
+    @property
+    def real(self: _Scalar) -> _Scalar: ...
```
shoyer: `real` and `imag` are not guaranteed to return the same scalar type -- they convert complex numbers to reals.

numpy/__init__.pyi
```diff
@@ -358,6 +358,187 @@ class ndarray(Iterable, Sized, SupportsInt, SupportsFloat, SupportsComplex,
...
+class bool_(generic): ...
+class object_(generic): ...
+class datetime64(generic): ...
+
+class integer(number): ...
+class signedinteger(integer): ...
+class int8(signedinteger): ...
+class int16(signedinteger): ...
+class int32(signedinteger): ...
+class int64(signedinteger): ...
```
eric-wieser: Sadly this doesn't actually reflect the underlying model - in reality, the types are `byte`, `short`, `int`, `long`, and `longlong`, and the numeric aliases are attached in a platform-dependent manner.

numpy/__init__.pyi
```diff
@@ -358,6 +358,187 @@ class ndarray(Iterable, Sized, SupportsInt, SupportsFloat, SupportsComplex,
...
+class uint32(unsignedinteger): ...
+class uint64(unsignedinteger): ...
+uint_ = uint64
+
+class inexact(number): ...
+class floating(inexact): ...
+class float16(floating): ...
+class float32(floating): ...
+class float64(floating): ...
+class float128(floating): ...
```
eric-wieser: Similarly the actual types here are `half`, `single`, `double`, `longdouble`, and similarly for complex.

numpy/__init__.pyi
eric-wieser: This is the approach recommended by PEP484

shoyer: OK, in that case we should probably fix `_DtypeLike` :)

numpy/__init__.pyi
shoyer: Does it really matter whether we define aliases or the real classes here? I would actually be OK with doing only these aliases. For annotations (and clearly written code in general), it's best to avoid platform dependent types like `int`.

numpy/__init__.pyi
eric-wieser: To be clear, this is the recommended approach for typing `self`

shoyer: I was suggesting that `bound="generic"` could be written as `bound=generic`.

numpy/__init__.pyi
eric-wieser: Well since there are actually 5 underlying types, sometimes numpy returns a type that is not referred to by any of these four aliases (which is made worse by numpy/numpy#10151 making them look the same).

I think what you have there is an argument about removing the unsized types from numpy - which I think is valid. But I think the type annotations should reflect what actually happens, not how we want things to happen.

numpy/__init__.pyi
eric-wieser: Which is relevant because `float128` is sometimes called `float96`, but both are always aliases for `longdouble`

numpy/__init__.pyi
eric-wieser: I realize. I suppose the PEP self-typing example is not within a stub file, so perhaps my point is irrelevant.

JelleZijlstra: In typeshed we prefer not doing any quoting of this sort.

alanhdu:
I've factored out an `_ArrayLike` class (although I need to play around with the types some more to see how to get it to typecheck the way I want).

I've also moved the platform-specific types together. AFAICT though, on the Python side the platform-specific types are aliases for the sized types (e.g. for me, `np.short.__name__ == 'int16'`), so the problem reduces to how to decide which sized type the unsized types correspond to (`mypy` supports `sys.platform`, but I'm not sure it supports `sys.maxsize`)).

eric-wieser:
> on the Python side the platform-specific types are aliases for the sized types

The `__name__` attributes are misleading
```
>>> np.int_.__name__  # on windows, this gives int32, so use `intc` below
'int64'
>>> np.longlong.__name__
'int64''
>>> np.int_ is np.longlong  # C long and C longlong are different types
False
```
There really is a unique type for each of the underlying `C` types. but the python-visible names are the sized name for that type. But the

alanhdu:
I believe that MyPy uses the actual type name (from a small test case):
```
Argument 1 to "f" has incompatible type "int64"; expected "longlong"
```

Is the "right" thing to do here is to just to have each of the platform specific types be their own distinct type? How important is it to get all the sized aliases matched up with their underlying components and how hard is that to do?

JelleZijlstra:
Mypy definitely doesn't look at the runtime `__name__` attribute; all it knows about numpy types necessarily comes from the stubs.

eric-wieser:
As part of `setup.py`, we could `import numpy` and generate the alias stubs based on the actual alias values

alanhdu:
I really like generating the alias stubs when the user installs them! I've opened up a follow-up issue (#15) to discuss that idea further.

For now, I've fixed the type errors and decided to leave the platform-specific types for another PR (which by default leaves them as `Any` types). Do you mind taking another look @shoyer @eric-wieser?

numpy/__init__.pyi
```diff
@@ -195,35 +195,24 @@ class flatiter:
...
 
 
-class ndarray(Iterable, Sized, SupportsInt, SupportsFloat, SupportsComplex,
-              SupportsBytes, SupportsAbs[Any]):
-
-    imag: ndarray
-    real: ndarray
-
+_ArraySelf = TypeVar("_ArraySelf", bound=_ArrayLike)
+class _ArrayLike(SupportsInt, SupportsFloat, SupportsComplex, SupportsBytes,
```
eric-wieser: Not a good name - in the docs, `array_like` means `can be converted to array`.

Perhaps `_ArrayOrScalarCommon`?

numpy/__init__.pyi
```diff
@@ -349,15 +316,112 @@ class ndarray(Iterable, Sized, SupportsInt, SupportsFloat, SupportsComplex,
...
+    def __getitem__(self, key) -> Any: ...
+    def __setitem__(self, key, value): ...
+    def __iter__(self) -> Any: ...
+    def __contains__(self, key) -> bool: ...
+    def __index__(self) -> int: ...
+
+class generic(_ArrayLike):
+    @property
+    def base(self) -> None: ...
+class _is_real(generic):
```
eric-wieser: Missing newline above `_is_real`, I think

eric-wieser: `_real_generic` would be a better name

numpy/__init__.pyi
```diff
@@ -349,15 +316,112 @@ class ndarray(Iterable, Sized, SupportsInt, SupportsFloat, SupportsComplex,
...
+    def imag(self) -> float32: ...
+class complex128(complexfloating):
+    @property
+    def real(self) -> float64: ...
+    @property
+    def imag(self) -> float64: ...
+
+class flexible(_is_real): ...
+class void(flexible): ...
+class character(_is_real): ...
```
eric-wieser: This is correct, but it's also bizarre:
```
>>> np.bytes_(b'test').real
'test'
>>> np.bytes_(b'test').imag
''
```

What were we thinking here?

alanhdu: If I had to guess, I think `imag` by default returns the "0" version of the data type -- while this makes sense for numbers, it has bizarre consequences for things that aren't numbers:

```python
In [4]: np.datetime64(dt.datetime.now()).real
Out[4]: numpy.datetime64('2018-03-26T15:22:14.672804')

In [5]: np.datetime64(dt.datetime.now()).imag
Out[5]: numpy.datetime64('1970-01-01T00:00:00.000000')
```

or 

```python
In [13]: np.object_(["hello world"]).real
Out[13]: array(['hello world'], dtype=object)

In [14]: np.object_(["hello world"]).imag
Out[14]: array([0], dtype=object)
```

shoyer:
👍 this looks great to me!

Two things that would be nice to have:
- Update the return value of `dtype.type` to `Type[generic]`.
- If you can think of any, add  a few additional smoke tests to our test file (e.g., verify that you can construct a numpy scalar and check its properties without an error from mypy)

numpy/__init__.pyi
eric-wieser: Issue opened at https://github.com/numpy/numpy/issues/10818

alanhdu:
@shoyer @eric-wieser I went a little overboard with the test suite and ended up writing a small test framework (I wanted to assert that certain things *failed* and that led me down a bit of a rabbit hole).

I put the documentation in the testing readme, the important bit is that:

There are three main directories of tests right now:
> There are three main directories of tests right now:
> 
> - `pass/` which contain Python files that must pass `mypy` checking with
>   no type errors
> - `fail/` which contain Python files that must *fail* `mypy` checking
>   with the annotated errors
> - `reveal/` which contain Python files that must output the correct
>   types with `reveal_type`
> 
> `fail` and `reveal` are annotated with comments that specify what error
> `mypy` threw and what type should be revealed respectively. The format
> looks like:
> 
> ```python
> bad_function   # E: <error message>
> reveal_type(x)   # E: <type name>
> ```
> 
> Right now, the error messages and types are must be **contained within
> corresponding mypy message**.
> 
> ## Running the tests
> 
> We use `py.test` to orchestrate our tests. You can just run:
> 
> ```
> py.test
> ```
> 
> to run the entire test suite. To run `mypy` on a specific file (which
> can be useful for debugging), you can also run:
> 
> ```
> $ cd tests
> $ MYPYPATH=.. mypy <file_path>
> ```

Is that ok? If not, I'm happy to move the test suite stuff to another PR for a longer discussion.

shoyer:
The test suite bit looks pretty sweet to me! Thanks @alanhdu

tests/test_stubs.py
```diff
@@ -0,0 +1,77 @@
...
+        if " E:" not in line and lineno not in errors:
+            continue
+
+        target_line = lines[lineno - 1]
+        if "# E:" in target_line:
+            marker = target_line.split("# E:")[-1].strip()
+            assert lineno in errors, f'Extra error "{marker}"'
+            assert marker in errors[lineno]
+        else:
+            assert "# E:" in target_line, f'Error "{errors[lineno]}" not found'
```
shoyer: Instead of the redundant assert, can you just use `pytest.fail()`?

shoyer: To guarantee quotes are escaped properly, let's cast with `repr()` inside the f-string, e.g., `f'Error {repr(errors[lineno])}" not found'`.

shoyer:
OK, merged. Thanks @alanhdu

PR #18, opened 2018-04-13
Update to be PEP 561 compliant

ethanhs:
This updates the stub package to be PEP 561 compliant, as a few things have changed since this repository was created For example, stub packages should have a suffix `-stubs` and stub packages don't need a py.typed file.

Also this uses the new searching released in mypy 0.590.


setup.py
```diff
@@ -1,4 +1,15 @@
-from setuptools import setup, find_packages
+from setuptools import setup
+import os
+
+
+def find_stubs(package):
```
shoyer: Just out of curiosity, will this sort of helper function be generally recommended for writing stub packages?

README.md
```diff
@@ -5,7 +5,7 @@
 This repository exists for developing [PEP 484](https://www.python.org/dev/peps/pep-0484/)
```
shoyer: Could you also kindly update `numpy_stubs` to `numpy-stubs` on the first line of the readme?

setup.py
ethanhs: Probably. I was thinking of making a change to either distutils or setuptools with helpers.

PR #19, opened 2018-05-08
Ensure stubs are valid for Python 2 and fix running of tests

FuegoFro:
The stubs contained an unconditional reference to SupportsBytes, which only exists in Python 3. To make these valid on Python 2, conditionally import that Protocol in Python 3 and otherwise use a dummy class in Python 2. Also have `ndarray` extend `Contains`, while we're here.

This also extends the test suites to run all tests against both Python 2 and Python 3, with the ability to specify that certain tests should only be run against Python 3 (eg to test Python 3 exclusive operators). This should help prevent errors like this moving forward.

One downside of this is that flake8 doesn't understand the `# type:` comments, so it thinks that imports from `typing` are unused. A workaround for this is to add `# noqa: F401` at the end of the relevant imports, though this is a bit tedious.

Finally, change how test requirements are installed and how the `numpy-stubs` package is exposed to mypy, and update the README/Travis file to reflect this. See https://github.com/python/mypy/issues/5007 for more details about the rational behind this change.

### DIFF
--- a/numpy-stubs/__init__.pyi
+++ b/numpy-stubs/__init__.pyi
@@ -1,14 +1,32 @@
 import builtins
-
-from typing import (
-    Any, Dict, Iterable, List, Optional, Mapping, Sequence, Sized,
-    SupportsInt, SupportsFloat, SupportsComplex, SupportsBytes, SupportsAbs,
-    Text, Tuple, Type, TypeVar, Union,
-)
-
 import sys
 
 from numpy.core._internal import _ctypes
+from typing import (
+    Any,
+    Container,
+    Dict,
+    Iterable,
+    List,
+    Mapping,
+    Optional,
+    Sequence,
+    Sized,
+    SupportsAbs,
+    SupportsComplex,
+    SupportsFloat,
+    SupportsInt,
+    Text,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+)
+
+if sys.version_info[0] < 3:
+    class SupportsBytes: ...
+else:
+    from typing import SupportsBytes
 
 _Shape = Tuple[int, ...]
 
@@ -325,7 +343,7 @@ class _ArrayOrScalarCommon(SupportsInt, SupportsFloat, SupportsComplex,
     def __getattr__(self, name) -> Any: ...
 
 
-class ndarray(_ArrayOrScalarCommon, Iterable, Sized):
+class ndarray(_ArrayOrScalarCommon, Iterable, Sized, Container):
     real: ndarray
     imag: ndarray
 
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -2,6 +2,3 @@ flake8==3.3.0
 flake8-pyi==17.3.0
 pytest==3.4.2
 mypy==0.590
-# This makes sure that the repo's stubs are accessible. Using MYPYPATH won't
-# work. See https://github.com/python/mypy/issues/5007 for more details.
-.
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -2,3 +2,6 @@ flake8==3.3.0
 flake8-pyi==17.3.0
 pytest==3.4.2
 mypy==0.590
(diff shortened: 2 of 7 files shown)


test-requirements.txt
```diff
@@ -2,3 +2,6 @@ flake8==3.3.0
 flake8-pyi==17.3.0
 pytest==3.4.2
 mypy==0.590
+# This makes sure that the repo's stubs are accessible. Using MYPYPATH won't
+# work. See https://github.com/python/mypy/issues/5007 for more details.
+.
```
shoyer: Interesting, I didn't realize that this worked in a requirements.txt file.

tests/pass/simple_py3.py
```diff
@@ -0,0 +1,6 @@
+import numpy as np
+
+array = np.array([1, 2])
+
+# The @ operator is not in python 2
+array @ array
```
ethanhs: Some people on 3.4 may want to run the test suite, so perhaps this shouldn't be used? Mypy defaults to using the python version of the executable used to run it.

test-requirements.txt
ethanhs: I prefer to keep the self install separate, as requirements are usually for "dependencies", but I suppose that is more of a stylistic nit.

numpy-stubs/__init__.pyi
```diff
@@ -1,14 +1,32 @@
...
+    Union,
 )
 
-import sys
-
-from numpy.core._internal import _ctypes
+if sys.version_info[0] < 3:
+    class SupportsBytes: ...
+else:
+    from typing import SupportsBytes
```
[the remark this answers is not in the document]
FuegoFro: I'm happy to pull this out to a separate commit if you'd like :) In my day job we only do squash-and-merge, so I'm not used to thinking about having granular commits *within* a PR haha.

test-requirements.txt
FuegoFro: @shoyer Yeah as far as I can tell, anything you can put after `pip install` you can put as a line in a requirements file :)

@ethanhs That's fair. I mainly wanted to simplify the process of getting the tests to work. But we can definitely remove this and change the README to tell people to run both commands
```
pip install -r test-requirements.txt
pip install .
```
if that's what folks prefer :)

tests/pass/simple_py3.py
FuegoFro: We could try to come up with a more generalized way of specifying what versions of python a given file can run on, or extend the mechanism so we have something like `_py<min_version>.py` files, where `min_version` could be 3, but could also be something like 3.5. If you think that's worth it I'm happy to do that :)

test-requirements.txt
shoyer: Let's do the two command version, just so it's more obvious what is going on. It's pretty non-standard to put `.` in a requirements.txt file.

numpy-stubs/__init__.pyi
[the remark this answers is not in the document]
shoyer: I've been doing squash-and-merge for most of the PRs to numpy-stubs, so I think it's fine to keep this as is.

shoyer:
I should have mentioned this before, but could you kindly add Python 2.7 (and maybe Python 3.5) to our test matrix configuration for Travis-CI?

test-requirements.txt
FuegoFro: Sounds good, I separated it out and updated the Travis file an README.

shoyer:
> I'm not entirely sure what this means in this context. In particular, mypy itself should always be run on Python 3, and I'm not sure if we need to test that the mypy tool works on Python 3.5 (that's what mypy's tests are for, right?).

OK, I think I missed this about mypy. So nevermind then :).

At some point it would be interesting to run unit tests with [pytype](https://github.com/google/pytype), but that's a whole separate project.

shoyer:
Thanks @FuegoFro !

FuegoFro:
Thank *you*!

PR #21, opened 2018-11-12
Add array conversion methods to ndarray

sproshev:
Actual types are like:

```
def item(self, *args: int) -> T: ...
@overload
def itemset(self, __value: T): ...
@overload
def itemset(self, __item: _ShapeLike, __value: T): ...
def astype(self,
           dtype: _DtypeLike,
           order: str=...,
           casting: str=...,
           subok: bool=...,
           copy: bool=...) -> ndarray[dtype]: ...
def byteswap(self, inplace: bool=...) -> ndarray[T]: ...
def copy(self, order: str=...) -> ndarray[T]: ...
def view(self, dtype: Union[_DtypeLike, NdArrayOrMatrix]=..., type: NdArrayOrMatrix=...) -> NdArrayOrMatrix[dtype]: ...
def getfield(self, dtype: Union[_DtypeLike, str], offset: int=...) -> ndarray[dtype]: ...
```


numpy-stubs/__init__.pyi
```diff
@@ -368,6 +370,35 @@ class ndarray(_ArrayOrScalarCommon, Iterable, Sized, Container):
...
+               dtype: _DtypeLike,
+               order: str=...,
+               casting: str=...,
+               subok: bool=...,
+               copy: bool=...) -> ndarray: ...
+    def byteswap(self, inplace: bool=...) -> ndarray: ...
+    def copy(self, order: str=...) -> ndarray: ...
+    def view(self,
+             dtype: Union[_DtypeLike, Type]=...,
+             type: Any=...) -> ndarray: ...
```
shoyer: This should probably substitute `Type` -> `Type[ndarray]` and `Any` -> `Type[ndarray]`.

sproshev: Thank you! Fixed. Covered methods with tests

tests/reveal/ndarray_conversion.py
```diff
@@ -0,0 +1,58 @@
...
+# copy
+reveal_type(nd.copy())  # E: numpy.ndarray
+reveal_type(nd.copy("C"))  # E: numpy.ndarray
+
+# view
+reveal_type(nd.view())  # E: numpy.ndarray
+reveal_type(nd.view(np.int64))  # E: numpy.ndarray
+reveal_type(nd.view(dtype=np.int64))  # E: numpy.ndarray
+reveal_type(nd.view(np.int64, np.matrix))  # E: numpy.ndarray
+reveal_type(nd.view(type=np.matrix))  # E: numpy.ndarray
```
shoyer: In theory, I think this should be `numpy.matrix`.

Or we can could make a dummy subclass, if `np.matrix` isn't defined, e.g.,
```
class SubArray(np.ndarray):
    pass

reveal_type(nd.view(type=SubArray))  # E: SubArray
```

Does that actually work?

sproshev: Thank you for review! I've removed simple tests and fixed type hints for `view`.

PR #22, opened 2018-11-12
Take `flake8` config from `typeshed`

sproshev:
See
https://github.com/PyCQA/pyflakes/issues/320
https://github.com/PyCQA/pyflakes/pull/371
https://github.com/python/typeshed/blob/master/.flake8


shoyer:
thanks @sproshev

PR #24, opened 2018-11-13
Add item selection and manipulation methods to ndarray

sproshev:
Please don't merge until I add tests :)

### DIFF
--- a/numpy-stubs/__init__.pyi
+++ b/numpy-stubs/__init__.pyi
@@ -38,9 +38,6 @@ _Shape = Tuple[int, ...]
 # Anything that can be coerced to a shape tuple
 _ShapeLike = Union[int, Sequence[int]]
 
-# Anything that mentioned as `array_like ` in  documentation
-_ArrayLike = Union[_T, Sequence[_T]]
-
 _DtypeLikeNested = Any  # TODO: wait for support for recursive types
 
 # Anything that can be coerced into numpy.dtype.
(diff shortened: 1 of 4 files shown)


sproshev:
I'll update this PR tomorrow because I found some mistakes while writing tests,

numpy-stubs/__init__.pyi
```diff
@@ -387,6 +394,58 @@ class ndarray(_ArrayOrScalarCommon, Iterable, Sized, Container):
     def ravel(self, order: str=...) -> ndarray: ...
     def squeeze(self, axis: Union[int, Tuple[int, ...]]=...) -> ndarray: ...
 
+    # Item selection and manipulation
+    @overload
+    def take(self, indices: int, *, mode: str=...) -> Any: ...
+    @overload
+    def take(self, indices: int, axis: int, out: ndarray=..., mode: str=...) -> Any: ...
+    @overload
+    def take(self, indices: Sequence[int], axis: int=..., out: ndarray=..., mode: str=...) -> ndarray: ...
```
shoyer: `indices` can take array-like arguments, not just `Sequence`.

Does `np.ndarray` pass `Sequence` checks?

numpy-stubs/__init__.pyi
```diff
@@ -29,11 +30,17 @@ if sys.version_info[0] < 3:
...
+# Used to type hint values that could be in `ndarray`
+_T = TypeVar("_T")
+
 _Shape = Tuple[int, ...]
 
 # Anything that can be coerced to a shape tuple
 _ShapeLike = Union[int, Sequence[int]]
 
+# Anything that mentioned as `array_like ` in  documentation
+_ArrayLike = Union[_T, Sequence[_T]]
```
shoyer: I support the idea of clearly document NumPy's interface, but in practice, how does this specific annotation differ from `Any`?

I get the idea of trying to constrain to only "array-like" inputs, but keep in mind that for NumPy that means practically anything -- we can coerce almost any type of object to ndarray with `np.array()`.

numpy-stubs/__init__.pyi
sproshev: Oh! Great finding, `ndarray` has all `Sequence`'s methods so I'll specify `Sequence` as a super-class

numpy-stubs/__init__.pyi
sproshev: I was trying to prepare `ndarray` to become generic and limit passed argument to have same `dtype` as `self` but after a bit more testing I agree with you. It's better to put `Any` here.

person142:

PR #26, opened 2018-11-15
Update ndarray's special methods

sproshev:


numpy-stubs/__init__.pyi
```diff
@@ -215,8 +208,7 @@ class flatiter:
 
 
 _ArraySelf = TypeVar("_ArraySelf", bound=_ArrayOrScalarCommon)
-class _ArrayOrScalarCommon(SupportsInt, SupportsFloat, SupportsComplex,
```
shoyer: Did something change to explain why there's no longer a point in subclassing from `SupportsInt`, etc?

JelleZijlstra: I think protocols make this subclassing unnecessary.

person142:

PR #27, opened 2018-11-15
 Add methods for calculation on ndarray 

sproshev:


numpy-stubs/__init__.pyi
```diff
@@ -387,6 +393,84 @@ class ndarray(_ArrayOrScalarCommon, Iterable, Sized, Container):
     def ravel(self, order: str=...) -> ndarray: ...
     def squeeze(self, axis: Union[int, Tuple[int, ...]]=...) -> ndarray: ...
 
+    # Calculation
+    @overload
+    def argmax(self) -> Any: ...
+    @overload
+    def argmax(self, axis: int, out: ndarray=...) -> Any: ...
```
shoyer: I don't quite understand -- why is it useful to define `argmax` twice? How is this different from simply defining a single signature where `axis` is optional?

sproshev: In this and other similar cases I'm trying to forbid using `out` when result is one element. I was unable to pass such `out` parameter that `argmax(out=np.empty(...))` does not raise an error.

shoyer: You can use `out` by passing in a 0-dimensional array, e.g.,
```
x = np.arange(5)
out = np.array(0)
x.argmax(out=out)
print(repr(out))
```
outputs: `array(4)`

person142:

PR #28, opened 2018-11-20
Bumpy mypy versions

alanhdu:
Bumping the `mypy` version required re-ordering some of the `overload`.

Bumping the `flake8` version created new errors, so I used https://github.com/ambv/black to autoformat the `*.pyi` stub files. (This is also helpful because it's what i used for the autogenerated stubs for #11). Is this the right approach, or should we just manually format these stub files?


.travis.yml
```diff
@@ -1,5 +1,7 @@
 language: python
-python: 3.6
+python:
+- 3.6
+- 3.7
```
shoyer: It looks like you may need to copy a few lines from NumPy's `.travis.yml` to check Python 3.7 working:
https://github.com/numpy/numpy/blob/4f1541e1cb68beb3049a21cbdec6e3d30c2afbbb/.travis.yml#L39-L41

shoyer:
I do like the look of `black` formatted code here. It's not what we normally do for NumPy but I don't see any harm in trying it here. @sproshev any thoughts?

max-sixty:
> I do like the look of black formatted code here. It's not what we normally do for NumPy but I don't see any harm in trying it here

FYI we've used it in `pandas-gbq`, it's been great for not worrying about formatting. We've also rolled it out to our smaller / newer internal repos. Would be a mammoth diff for a project like xarray though

shoyer:
Thanks!

PR #29, opened 2019-06-07
Added hint for ndarray.__new__

bedder:
`numpy.ndarray.__new__` previously did not have any type definitions, meaning that code that directly instantiates `np.ndarray` instances would be flagged by mypy as having unexpected arguments.

I think this should fix the issue - at the very least this change removes what I consider to be invalid warnings on my codebase - although I'm unsure about whether the type for the buffer is correct.

### DIFF
--- a/numpy-stubs/__init__.pyi
+++ b/numpy-stubs/__init__.pyi
@@ -278,9 +278,20 @@ class _ArrayOrScalarCommon(
     # TODO(shoyer): remove when all methods are defined
     def __getattr__(self, name) -> Any: ...
 
+_BufferType = Union[ndarray, bytes, bytearray, memoryview]
+
 class ndarray(_ArrayOrScalarCommon, Iterable, Sized, Container):
     real: ndarray
     imag: ndarray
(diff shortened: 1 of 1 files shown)


shoyer:
There seems to be a bigger issue about defining buffer types (https://github.com/python/typing/issues/593), but this seems like a good start.

Thanks @bedder !

bedder:
Thanks to everyone working on this project - it's a really useful addition to numpy!

At my work we've ended up creating a [public fork of this project](https://github.com/Prowler-io/numpy-stubs), and have slowly been adding type hints for more of the numpy library. If it would be useful to yourselves (and not too spammy), we'd be happy to raise more PRs for the additions we've been using internally.

shoyer:
Yes, we would be very happy to review pull requests!

shoyer:
Feel free to ping me directly for reviews -- this repo has been a little
quiet recently (and GitHub's notifications UI for maintainers is a bit of a
mess) so I have been missing some activity here.

PR #31, opened 2019-07-09
Added typing information for datetime64, timedelta64

bedder:
Additions to support the `datetime64` and `timedelta64` classes, as described [here](https://docs.scipy.org/doc/numpy/reference/arrays.datetime.html).


numpy-stubs/__init__.pyi
```diff
@@ -393,14 +394,39 @@ class number(generic):
 
 class bool_(_real_generic): ...
 class object_(generic): ...
-class datetime64(_real_generic): ...
+
+class datetime64:
```
bedder: Motivation behind removing inheritance of `_real_generic`: inheritance allowed code like this to pass `mypy`
```
np.timedelta64(0) - np.datetime64(0, 'D')
```

tests/test_stubs.py
```diff
@@ -50,14 +51,14 @@ def test_fail(path, py2_arg):
     with open(path) as fin:
         lines = fin.readlines()
 
-    errors = {}
+    errors = defaultdict(lambda: "")
```
bedder: The previous iteration of the code didn't work too well when we had multi-line error messages (e.g. with overload matching errors), as it would just overwrite the error messages.

shoyer:
Thanks @bedder!

PR #33, opened 2019-09-17
Array Creation Routines: zeros and ones

szb0:
I added a couple of commonly used array creation routines (ones and zeros).

One thing that I would imagine is out of scope of this PR and likely out of scope of this project is pythons treatment of `bool` as a subclass of `int`. This means with `_ShapeLike` we can have something like `np.zeros(False)` pass the analysis, while numpy at runtime raises a `TypeError`.


shoyer:
This looks great, thank you!

Issue #15, opened 2018-03-25
Use a stub-generator


alanhdu:
I think we should seriously consider having a build system that generates the actual NumPy stubs. @eric-wieser suggested that for platform-specific types (e.g. `np.intp`) we:

> As part of setup.py, we could import numpy and generate the alias stubs based on the actual alias values

which seems like a nice and clean solution (especially for types like `np.float96` which may or may not exist depending on the platform). Adding some kind of "stub-generation" step would also let us reduce a lot of boilerplate that we'd have to do otherwise:

- All distribution generators in `numpy.random` should have the same type signatures as `numpy.random.RandomState` (so they need to be copy-pasted and kept in sync)
- The interactions between arrays and scalars and broadcasting sometimes require a lot of repetitive `@overload` functions (see the distributions in #11 as an example).

This would also give us a chance to enforce some lints of our own, like:
- The various `*Like` types should only ever be taken as function input types and never as return types.

Of course, there are also downsides with having a stub-generation step:
- Generating valid Python stubs can be tricky (due to significant whitespace) and there don't seem to be standard solutions to do this (AFAIK).
- This adds more complexity to the project -- instead of being able to just read the stubs and make changes to fix bugs people will have to understand whatever generator framework we use and debug that too
- If we run the stub generation when a user installs stubs, then whatever dependencies we use for framework generation will have to be pre-installed or vendored

I think that the added complexity would be worth it, but I'm curious what everyone else's opinions are.

cc @shoyer @eric-wieser

eric-wieser:
> If we run the stub generation when a user installs stubs, then whatever dependencies we use for framework generation will have to be pre-installed or vendored

Can't we use PEP518 and `pyproject.toml` here to deal with that? Looks like support was added in https://github.com/pypa/pip/pull/4144

eric-wieser:
> This would also give us a chance to enforce some lints of our own

This sounds more in scope for a mypy plugin

>  people will have to understand whatever generator framework we use

I might be inclined to pick something simpler like a templating framework

shoyer:
Yes, I'm generally supportive.

I can definitely see this making sense if/when we try to add support for dtypes. Most NumPy ufuncs only work for particular dtypes but we would go absolutely insane duplicating that without reference to the ufunc objects from NumPy.

alanhdu:
> I might be inclined to pick something simpler like a templating framework

[...]

PR #34, opened 2019-10-01
Add some operations to ndarray

kbrose:
I just went through the `ndarray` docs alphabetically and added some operations. I wanted to submit for review before I got too far -- is this an approach you would consider including, or are you looking to get all the operations typed in some other way (e.g. auto-generating per #15)?


kbrose:
Is this repo still being developed?

numpy-stubs/__init__.pyi
```diff
@@ -368,6 +368,58 @@ class ndarray(_ArrayOrScalarCommon, Iterable, Sized, Container):
     def flatten(self, order: str = ...) -> ndarray: ...
     def ravel(self, order: str = ...) -> ndarray: ...
     def squeeze(self, axis: Union[int, Tuple[int, ...]] = ...) -> ndarray: ...
+    # Operations
+    @overload
+    def all(
+        self, axis: None = ..., out: None = ..., keepdims: Optional[bool] = ...
+    ) -> bool_: ...
+    @overload
+    def all(
```
person142: Note that specifying an axis doesn't necessarily mean you will get an array out; e.g.

```python
>>> x = np.array([True, True])
>>> x.all(axis=0)
True
```

And the analogue with more axes:

```python
>>> x = np.ones((2, 2), dtype=np.bool_)
>>> x.all(axis=(0, 1))
True
```

If we had `Literal` we could make that guarantee for when `keepdims = True`, but we can't do that yet.

numpy-stubs/__init__.pyi
```diff
@@ -368,6 +368,58 @@ class ndarray(_ArrayOrScalarCommon, Iterable, Sized, Container):
     def flatten(self, order: str = ...) -> ndarray: ...
     def ravel(self, order: str = ...) -> ndarray: ...
     def squeeze(self, axis: Union[int, Tuple[int, ...]] = ...) -> ndarray: ...
+    # Operations
+    @overload
+    def all(
```
person142: You can specify a 0d array as the `out` even when not specifying an axis, e.g.

```python
>>> x = np.array([True, True])
>>> out = np.array(True, dtype=np.bool_)
>>> x.all(out=out)
array(True)
>>> out
array(True)
```

numpy-stubs/__init__.pyi
```diff
@@ -368,6 +368,58 @@ class ndarray(_ArrayOrScalarCommon, Iterable, Sized, Container):
...
+        self, axis: None = ..., out: None = ..., keepdims: Optional[bool] = ...
+    ) -> bool_: ...
+    @overload
+    def all(
+        self, axis: int, out: Optional[ndarray] = ..., keepdims: Optional[bool] = ...
+    ) -> ndarray: ...
+    @overload
+    def any(
```
person142: Similar comments here as with `all`.

numpy-stubs/__init__.pyi
```diff
@@ -368,6 +368,58 @@ class ndarray(_ArrayOrScalarCommon, Iterable, Sized, Container):
...
+        self, axis: None = ..., out: None = ..., keepdims: Optional[bool] = ...
+    ) -> bool_: ...
+    @overload
+    def any(
+        self, axis: int, out: Optional[ndarray] = ..., keepdims: Optional[bool] = ...
+    ) -> ndarray: ...
+    @overload
+    def argmax(self, axis: None = ..., out: Optional[ndarray] = ...) -> integer: ...
+    @overload
+    def argmax(self, axis: int, out: Optional[ndarray] = ...) -> ndarray: ...
```
person142: Similarly we might get a scalar output here:

```python
>>> x = np.arange(4)
>>> x.argmax(axis=0)
3
```

kbrose:
I'm no longer motivated to finish this at this point in time.

PR #38, opened 2019-11-24
Add type hinting for numeric.py

johanvergeer:


johanvergeer:
I have to add the tests for this file, but I first wanted to get your opinion.

numpy-stubs/__init__.pyi
```diff
@@ -82,6 +83,19 @@ _DtypeLike = Union[
 
 _NdArraySubClass = TypeVar("_NdArraySubClass", bound=ndarray)
 
+_ArrayLike = TypeVar(
+    "_ArrayLike",
+    str,
+    int,
+    float,
+    bool,
+    object,
```
shoyer: Is there a meaningful difference between `TypeVar("_ArrayLike")`,  `TypeVar("_ArrayLike", object)` and what you wrote here?

I agree that it's a good idea to clearly identify "array like" arguments, but I'm not certain this the right annotation. Note that in most cases NumPy is fine with absolutely *any* type of object in cases where it needs an "array like" argument

johanvergeer: You have a good point. I was already hesitating on how to create this TypeVar, but since an `array_like` can be just about anything, it makes better sense to just declare it as `_ArrayLike = TypeVar("_ArrayLike")`

Best thing is, we can always change this TypeVar if we come to another solution.

aldanor:
Any plans on merging this?

shoyer:
Apologies for the delays here. It would be great to get someone else who is excited about type checking with NumPy involved in maintaining this repository -- right now it is easy for notifications to get lost since activity is relatively low. Alternatively, maybe now is finally the time to move annotations into NumPy proper?

johanvergeer:
Hi @shoyer 

I didn't continue to work on it because of the slow responses. 
If we can start to add Mypy directly into Numpy I'm happy to help.

Issue #42, opened 2020-03-20
Type error with `np.dtype`


person142:
### Minimal reproducer

Running mypy on the following code:

```python
import numpy as np

object_dtype = [('field1', object)]
np.dtype(object_dtype)
```

Results in the following error:

```
$ mypy test_dtype.py
test_dtype.py:4: error: Argument 1 to "dtype" has incompatible type "List[Tuple[str, Type[object]]]"; expected "Union[dtype, None, type, str, Tuple[Any, int], Tuple[Any, Union[int, Sequence[int]]], List[Union[Tuple[Union[str, Tuple[str, str]], Any], Tuple[Union[str, Tuple[str, str]], Any, Union[int, Sequence[int]]]]], Dict[str, Union[Sequence[str], Sequence[Any], Sequence[int], Sequence[Union[bytes, str, None]], int]], Dict[str, Tuple[Any, int]], Tuple[Any, Any]]"
```

Note that putting `object_dtype` on the previous line is important; otherwise the context of the function signature:

https://mypy.readthedocs.io/en/latest/type_inference_and_annotations.html#context-in-type-inference

will broaden the type to `List[Tuple[str, Any]]` (I think.)

### Context

Stumbled across this while working on typing in SciPy, with the current version of the NumPy stubs there's an error because of this line:

https://github.com/scipy/scipy/blob/master/scipy/io/matlab/tests/test_mio.py#L140

### Speculation

I _think_ this is happening because `List` is invariant and `object` is a subtype of `Any`, so despite the fact that `_DtypeLikeNested` is typed as any here:

https://github.com/numpy/numpy-stubs/blob/master/numpy-stubs/__init__.pyi#L41

`object` still isn't let through here:

https://github.com/numpy/numpy-stubs/blob/master/numpy-stubs/__init__.pyi#L61

PR #43, opened 2020-03-21
MAINT: broaden allowed lists in `_DtypeLike`

person142:
Closes https://github.com/numpy/numpy-stubs/issues/42.

`np.dtype` accepts a wide variety of lists as dtypes, some examples
include:

- `[('R', 'u1'), ('G', np.unicode_, 1)]`
- `[('R', 'u1', (2, 2)), ('G', np.unicode_, 1)]`

Without further hints mypy is going to type those as `List[object]`,
so precise hints will require that users add annotations. In addition,
typing the above is challenging because the inputs have to be a list,
which is invariant, so you can't use `Union`s inside the list to
reduce the number of cases. Work around that by just accepting
`List[Any]`.

### DIFF
--- a/tests/pass/simple.py
+++ b/tests/pass/simple.py
@@ -25,9 +25,26 @@ np.dtype('float64')
 np.dtype(np.dtype(float))
 np.dtype(('U', 10))
 np.dtype((np.int32, (2, 2)))
-np.dtype([('R', 'u1'), ('G', 'u1'), ('B', 'u1')])
-np.dtype([('R', 'u1', 1)])
-np.dtype([('R', 'u1', (2, 2))])
+# Define the arguments on the previous line to prevent bidirectional
+# type inference in mypy from broadening the types.
+two_tuples_dtype = [('R', 'u1'), ('G', 'u1'), ('B', 'u1')]
+np.dtype(two_tuples_dtype)
+
+three_tuples_dtype = [('R', 'u1', 1)]
+np.dtype(three_tuples_dtype)
+
+mixed_tuples_dtype = [('R', 'u1'), ('G', np.unicode_, 1)]
+np.dtype(mixed_tuples_dtype)
+
+shape_tuple_dtype = [('R', 'u1', (2, 2))]
+np.dtype(shape_tuple_dtype)
+
+shape_like_dtype = [('R', 'u1', (2, 2)), ('G', np.unicode_, 1)]
+np.dtype(shape_like_dtype)
+
+object_dtype = [('field1', object)]
+np.dtype(object_dtype)
+
 np.dtype({'col1': ('U10', 0), 'col2': ('float32', 10)})
 np.dtype((np.int32, {'real': (np.int16, 0), 'imag': (np.int16, 2)}))
 np.dtype((np.int32, (np.int8, 4)))
--- a/numpy-stubs/__init__.pyi
+++ b/numpy-stubs/__init__.pyi
@@ -56,12 +56,11 @@ _DtypeLike = Union[
     # (fixed_dtype, shape)
     Tuple[_DtypeLikeNested, _ShapeLike],
     # [(field_name, field_dtype, field_shape), ...]
-    List[
-        Union[
-            Tuple[Union[str, Tuple[str, str]], _DtypeLikeNested],
-            Tuple[Union[str, Tuple[str, str]], _DtypeLikeNested, _ShapeLike],
-        ]
-    ],
+    #
+    # The type here is quite broad because NumPy accepts quite a wide
+    # range of inputs inside the list; see the tests for some
+    # examples.
+    List[Any],
     # {'names': ..., 'formats': ..., 'offsets': ..., 'titles': ...,
     #  'itemsize': ...}
     # TODO: use TypedDict when/if it's officially supported


tests/pass/simple.py
```diff
@@ -25,9 +25,26 @@ def ndarray_func(x):
...
 np.dtype((np.int32, (2, 2)))
-np.dtype([('R', 'u1'), ('G', 'u1'), ('B', 'u1')])
-np.dtype([('R', 'u1', 1)])
-np.dtype([('R', 'u1', (2, 2))])
+# Define the arguments on the previous line to prevent bidirectional
+# type inference in mypy from broadening the types.
+two_tuples_dtype = [('R', 'u1'), ('G', 'u1'), ('B', 'u1')]
+np.dtype(two_tuples_dtype)
+
+three_tuples_dtype = [('R', 'u1', 1)]
```
person142: Note that if you moved the dtype to the previous line to avoid broadening, this test would fail on master.

person142:
I don't love this solution-originally I was trying to make a giant enumeration of all the cases, but there are a lot of them. Could be handled better by some codegen. Even then, it doesn't solve the problem that for exotic cases users will have to add annotations; maybe that's ok though.

charris:
@seberg Ping.

seberg:
Well, you can do this I guess. There is another possibly broadening (not in general though maybe), that is to include `type` itself, for things like `np.array(..., dtype=np.string)` and new DType classes.

My wonder about this (and some other array-like constructs), is whether we should actually allow all things that are allowed? I can see an as broad as possible `_DTypeLike` as necessary for `np.dtype`, but I wonder if it is really not sensible to expect users to do `np.array(..., dtype=np.dtype(...))` to get typing if they want to use typing? There is the above other hangle, in that `np.dtype(np.floating)` does not make sense, but `dtype=np.floating` *can* make sense, so these two differ anyway IMO.

So the broadening makes sense to me (it is just correct for `np.dtype`), but I am wondering if you should not do the opposite of broading for anything that takes a `dtype` argument.

person142:
> that is to include type itself

If I am understanding your comment correctly, `type` is already included:

https://github.com/numpy/numpy-stubs/blob/d7e56a4948af158ac4b41d79b727dcaf131087fe/numpy-stubs/__init__.pyi#L50

(Note that the full definition of `_DtypeLike` is a giant union; I'm just editing one entry of that union to generalize what gets counted as "list like dtypes".)

person142:
> I wonder if it is really not sensible to expect users to do np.array(..., dtype=np.dtype(...)) to get typing if they want to use typing?

Indeed, and because of that the types allow for much more since the `dtype` argument to `np.array` is typed as `_DtypeLike`:

https://github.com/numpy/numpy-stubs/blob/master/numpy-stubs/__init__.pyi#L480

seberg:
Right about `type`. There is a difference in meaning, `np.float64` actually means the same as `"float64"`. `np.string` does mean the same as `np.dtype(np.string)` *but* that will probably not be true forever, `np.dtype(np.string)` might be an error in the future, while `np.array(..., dtype=np.string)` would not be. That is because `np.dtype(np.string)` is not actually an object that can be attached to a NumPy array itself.

I am not sure what you mean with your second message. Do you think the `dtype` argument should be more restrictive or arguing that it should include everything `np.dtype` does?

person142:
> Do you think the dtype argument should be more restrictive or arguing that it should include everything np.dtype does?

I was arguing that it should (and in the current types does) include everything `np.dtype` does, though I now see from your comment that this might need to be changed in the future.

seberg:
There are subtle differences, yeah. Although unless you type much more strictly (and add new ABCs or so), there is probably no way to distinguish these subtleties.

Anyway, if the idea is to be as broad as possible (I do not use types, although I am not sure I agree with the sentiment, since I feel the types should promote best practices). The broadening as such seems good, although I am not sure why you do not type it correctly? I am pretty sure you can add a `_DTypeTuple` type which is a union of a tuple with the 2 or 3 entries and then say it is a list of said tuple? That seems like it gives the same result, without allowing also letting pass fully incorrect things.

seberg:
Oh, sorry, so that a union inside the list does not work? You could still use at least a list of tuples probably.

Anyway, I would prefer someone who uses type annotations more to chip in, whether they want to broaden it up or what should the API should be.

person142:
> Oh, sorry, so that a union inside the list does not work? You could still use at least a list of tuples probably.

Yup, that's the problem. And a list of tuples has the same issues. An example of that not involving dtypes at all:

```python
from typing import Any, List, Tuple

x = [(1, 2)]

def f(x: List[Tuple]):
    pass

# Actually the same as the first example.
def g(x: List[Tuple[Any, ...]]):
    pass

def h(x: List[Tuple[int, ...]]):
    pass

f(x)
g(x)
h(x)
```

And the mypy output:

```
tuples.py:15: error: Argument 1 to "f" has incompatible type "List[Tuple[int, int]]"; expected "List[Tuple[Any, ...]]"
tuples.py:15: note: "List" is invariant -- see http://mypy.readthedocs.io/en/latest/common_issues.html#variance
tuples.py:15: note: Consider using "Sequence" instead, which is covariant
tuples.py:16: error: Argument 1 to "g" has incompatible type "List[Tuple[int, int]]"; expected "List[Tuple[Any, ...]]"
tuples.py:16: note: "List" is invariant -- see http://mypy.readthedocs.io/en/latest/common_issues.html#variance
tuples.py:16: note: Consider using "Sequence" instead, which is covariant
tuples.py:17: error: Argument 1 to "h" has incompatible type "List[Tuple[int, int]]"; expected "List[Tuple[int, ...]]"
tuples.py:17: note: "List" is invariant -- see http://mypy.readthedocs.io/en/latest/common_issues.html#variance
tuples.py:17: note: Consider using "Sequence" instead, which is covariant
```

In that example you could fix things by adding an explicit annotation to `x`:

```
x: List[Tuple] = [(1, 2)]
```

If desired we could certainly do something here like `List[Tuple]`, and then places like

https://github.com/scipy/scipy/blob/master/scipy/io/matlab/tests/test_mio.py#L139

will have to add the annotation `List[Tuple]` to pass mypy (seems fine). You could of course take that even further and require more complex annotations, but they would have to be something like `List[Tuple[str, Any]]` (since `_DtypeLikeNested` is `Any`), which would probably be pretty surprising for the end user (so maybe not good to take it that far).

ethanhs:
You may consider using `Sequence[T]` instead of `List[T]` for arguments, as Sequence is not invariant.

https://mypy-play.net/?mypy=latest&python=3.8&gist=164acb8d6bcbcfae3a83e4d4c9c1d47f

person142:
Alas we can’t use `Sequence` as NumPy literally only accepts a list here. (You can test that with a tuple of tuples-it will fail.)

seberg:
@shoyer maybe a ping in case you did not see this (not that I am sure you want to look at it). Since I do not use these, I am not sure I should be complaining, but I am still unsure that I do not think strict typing is good. Just like `List` seems typed as homogeneous which is a huge limitation from typical lists. You could for example also type it `List[(str, dtype)]` and force users to write:
```
field1 = "name1", np.dtype((np.int8, (20,)))
field2 = "name2", np.dtype("S8")
np.dtype([field1, field2])
```
Yeah, it is ugly, but such structured DTypes and the whole "mini-language" around it is annoying IMO. On the other hand... I should probably just not bother, at some point I might rather like to have a "helper" type `np.structured(...)` with a clearer signature.

shoyer:
Thanks for the ping, I missed this. I agree, this seems like a practical improvement to me.

PR #44, opened 2020-03-22
ENH: add basic typing for `np.ufunc`

person142:
This adds basic type hints for `np.ufunc` and types all ufuncs in the
top-level namespace as `np.ufunc`. Ufuncs are highly dynamic, e.g.

- Their call signatures can vary
- The `reduce`/`accumulate`/`reduceat`/`outer`/`at` methods may or may
  not always raise

so it is difficult to have precise types. A path forward would be
writing a mypy plugin to get more precise typing.


scripts/find_ufuncs.py
```diff
@@ -0,0 +1,21 @@
+import numpy as np
+
+
+def main():
```
person142: This is the script I used to generate the big list of ufuncs. Could ultimately be done as part of a build step.

numpy-stubs/__init__.pyi
```diff
@@ -489,5 +490,160 @@ def ones(
...
+        # TODO: make this precise when we can use Literal.
+        order: Optional[str] = ...,
+        dtype: Optional[_DtypeLike] = ...,
+        subok: bool = ...,
+        signature: Union[str, Tuple[str]] = ...,
+        # In reality this should be a length of list 3 containing an
+        # int, an int, and a callable, but there's no way to express
+        # that.
+        extobj: List[Union[int, Callable]] = ...,
+    ) -> Union[ndarray, generic]: ...
```
person142: I think that after a mypy upgrade to resolve this issue

https://github.com/python/typing/issues/253

the union can be removed and replaced with two overloads (one for all scalars and one for array like). On the pinned version of mypy that doesn't work because the signatures have overlapping input types.

shoyer:
Thanks!

PR #45, opened 2020-03-26
MAINT: upgrade mypy to 0.770 (the latest)

person142:
The most significant benefit is changes to `@overload` around what
counts as overlapping signatures that could unlock some functionality;
see e.g.

- https://github.com/numpy/numpy-stubs/pull/44#discussion_r396186738
- https://github.com/numpy/numpy-stubs/pull/11#issuecomment-369444243

### DIFF
--- a/tests/test_stubs.py
+++ b/tests/test_stubs.py
@@ -1,4 +1,5 @@
 import os
+import re
 from collections import defaultdict
 
 import pytest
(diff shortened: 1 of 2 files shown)


tests/test_stubs.py
```diff
@@ -52,7 +56,12 @@ def test_fail(path, py2_arg):
         lines = fin.readlines()
 
     errors = defaultdict(lambda: "")
-    for error_line in stdout.split("\n"):
+    error_lines = stdout.rstrip("\n").split("\n")
+    assert re.match(
```
person142: Now mypy always outputs at least one summary line; check that this line is as expected (in case it goes away in a later version) and then drop it.

person142:
This is a straightforward change, so I will merge it in a day or two unless someone complains.

charris:
Might be worth adding a dependabot file at some point to track mypy releases.

rgommers:

Issue #7, opened 2017-12-12
Support for dtypes


shoyer:
Per https://github.com/numpy/numpy/issues/7370#issuecomment-346126667, we could specify dtype as part of ndarray types (including type inference) by using NumPy's [scalar types](https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.scalars.html), e.g., `np.ndarray[Any, np.float64]` or `np.ndarray[np.float64, Any]` (where `Any` indicates the shape).

Is this a good way to indicate dtypes? How would we handle dtypes that don't have a well specified scalar type? For example:
- `datetime64` and `timedelta64` do not have precision as part of the type
- [`np.void`](https://stackoverflow.com/questions/25247190/numpy-void-type-how-to-use-it) is used for all arbitrary structured dtypes, but doesn't indicate the field names or subtypes (side note: `np.struct` would be a much better name)
- string dtypes have their size as part of the type

One simple answer is to not worry about more detailed type information for now, and worry about more precise specification for scalar types later, when support in the typing module becomes available (e.g., for integer types or custom struct-like data structures like `TypedDict`).

rmcgibbo:
I think `ndarray[ShapeType, DtypeType]` is the best way to so specify the type.

My reason for preferring that the shape specification come before the dtype specification is that I anticipate that when things get a little verbose, we'll probably want to have type aliases, and putting the `ShapeType` first in the syntax makes something like a `array1d[Dtype]` as a shorthand (that looks like a partially applied function) for `ndarray["OneDimensional", Dtype]` makes much more sense than `arrayf64[ShapeType]` as a shorthand for `ndarray[ShapeType, float64]`.

rmcgibbo:
For  complex dtypes like `datetime64`, `np.void`, and the string dtypes, I propose that we delay dealing with them until this package is much more fully developed.

Very little functionality is lost in the meantime if numpy functions that return arrays of `dtype('S1')` are labeled as returning (or are inferred to return) results of type `ndarray[Any, str]`. More specificity can be added later.

ryanpeach:
This is really all I need, what is the issue? Anything I can do?

person142:
Circling back around to this: another issue is that you can mutate an array's dtype in-place, e.g.

```python
>>> x = np.ones((2,), dtype=np.int64)
>>> x
array([1, 1])
>>> x.dtype = np.bool_
>>> x
array([ True, False, False, False, False, False, False, False,  True,
[...]

PR #47, opened 2020-03-29
MAINT: ban setting `ndarray.dtype`

person142:
See discussion in https://github.com/numpy/numpy-stubs/issues/7.

Though this is valid NumPy, users should be using `ndarray.view`
instead of setting the dtype, and setting the dtype prevents sensibly
making `ndarray` generic over dtype because of situations like this:

```
x = np.ones((2,), dtype=np.int64)  # type is ndarray[np.int64]
x.dtype = np.bool_  # What is the type now?
```

If someone really wants to do this, they will now have add an ignore,
which should make it clear that type safety is going out the window.


shoyer:
Thanks!

PR #49, opened 2020-04-04
TST: add script to run tests

person142:
Since stubs-only packages have to be installed, add a simple script that installs the stubs and then invokes pytest.

Just a small shortcut for

```
$ pip install .
$ pytest
```


person142:
Since this is a small PR I'll merge in a day or so barring complaints.

rgommers:
LGTM, except shouldn't the filename be `runtests.py` rather than `runtests`?

person142:
> except shouldn't the filename be runtests.py rather than runtests?

Ah yeah, not good to buck conventions. Changed it.

runtests.py
```diff
@@ -0,0 +1,21 @@
...
+import sys
+
+import pytest
+
+STUBS_ROOT = os.path.dirname(os.path.abspath(__file__))
+
+
+def main():
+    subprocess.run(
+        ['pip', 'install', STUBS_ROOT],
```
rgommers, line 13: Oh now the more substantial comment: I don't think this works correctly in a virtualenv or conda env. It may install into the base env instead.

rgommers: Can depend on OS and env vars, see e.g. https://stackoverflow.com/questions/8052926/running-subprocess-within-different-virtualenv-with-python

person142: Hm ok, switched it to do `[sys.executable, '-m', 'pip', ...]`. It seems like if we were trying to do something more complicated we could still run into problems, but since we're just invoking pip we're ok?

rgommers: That should work I think.

rgommers:
Merged, thanks @person142

PR #50, opened 2020-04-11
MAINT: move things in `numeric.pyi` to the top-level `__init__.pyi`

person142:
In gh-38 some stubs were added for `core/numeric.py`. The problem is
that these stubs are not picked up at all, as can be tested with

```python
import numpy as np

reveal_type(np.ones_like)
```

which on master gives

```
Revealed type is 'Any'
```

The issue is that the NumPy main repo importing from numeric doesn't
automatically mean the type stubs do. We could add the relevant
imports on the stubs level too, but there is no need for the structure
of the stubs to mimic the structure of NumPy unless they are being
used to type-check NumPy itself (which we aren't doing yet), and
indeed having them mimic the internal structure just causes extra pain
of trying to keep up with internal refactors that don't affect the
public API. In fact, we already ignore NumPy's internal structure by
defining `ndarray` in the top-level `__init__.pyi`, when it is
actually exported by `numeric`.

So instead, just fix the issue by moving the numeric stubs into the
top level `__init__.pyi` (plus some minor bug fixes). On master the
above script now correctly gives:

```
Revealed type is 'def [_ArrayLike] (a: _ArrayLike`-1, dtype: Union[numpy.dtype, None] =, order: builtins.str =, subok: builtins.bool =, shape: Union[builtins.int, typing.Sequence[builtins.int], None] =) -> numpy.ndarray
````

### CODE REVIEW

**Comment by person142** [comment]
Note that these functions still have no tests (which is how this didn't get caught originally); I didn't try to fix that in this PR.

File: numpy-stubs/core/numeric.pyi [hunk 1]
```diff
@@ -1,111 +0,0 @@
...
-def isclose(
-    a: _ArrayLike,
-    b: _ArrayLike,
-    rtol: float = ...,
-    atol: float = ...,
-    equal_nan: bool = ...,
-) -> _ArrayLike: ...
-def array_equal(a1: _ArrayLike, a2: _ArrayLike) -> bool: ...
-def array_equiv(a1: _ArrayLike, a2: _ArrayLike) -> bool: ...
-def extend_all(module: ModuleType): ...
```
person142, line 111: This isn't a public function, so I removed it.

numpy-stubs/core/numeric.pyi
```diff
@@ -1,111 +0,0 @@
...
-        int, Tuple[int, int], Tuple[Tuple[int, int], ...], Tuple[List[int, int], ...]
-    ] = ...,
-) -> ndarray: ...
-def roll(
-    a: _ArrayLike,
-    shift: Union[int, Tuple[int, ...]],
-    axis: Optional[Union[int, Tuple[int, ...]]] = ...,
-) -> _T: ...
-def rollaxis(a: _ArrayLike, axis: int, start: int = ...) -> ndarray: ...
-def normalize_axis_tuple(
```
person142, line 68: Not a public function; removed it.

numpy-stubs/core/numeric.pyi
```diff
@@ -1,111 +0,0 @@
...
-_T = TypeVar("_T")
-
-def zeros_like(
-    a: _ArrayLike,
-    dtype: Optional[dtype] = ...,
-    order: str = ...,
-    subok: bool = ...,
-    shape: Optional[Union[int, Sequence[int]]] = ...,
-) -> ndarray[int]: ...
-def ones(
```
person142, line 25: Already defined in `__init__.pyi`.

numpy-stubs/core/numeric.pyi
```diff
@@ -1,111 +0,0 @@
...
-
-_T = TypeVar("_T")
-
-def zeros_like(
-    a: _ArrayLike,
-    dtype: Optional[dtype] = ...,
-    order: str = ...,
-    subok: bool = ...,
-    shape: Optional[Union[int, Sequence[int]]] = ...,
-) -> ndarray[int]: ...
```
person142, line 24: Converted the various `ndarray[T]` things to `ndarray`, since that's not a thing yet (https://github.com/numpy/numpy-stubs/pull/48).

rgommers:
> So instead, just fix the issue by moving the numeric stubs into the
> top level `__init__.pyi` (plus some minor bug fixes).

makes sense to me

person142:
Ok, fixed the pyflakes error.

rgommers:

PR #51, opened 2020-04-15
ENH: add numerical constants

person142:
Scraped them with this script:

```python
import numpy as np


def main():
    constants = []
    for obj_name in dir(np):
        obj = getattr(np, obj_name)
        if isinstance(obj, float):
            constants.append(f'{obj_name}: float')
    print('\n'.join(sorted(constants)))


if __name__ == '__main__':
    main()
```


rgommers:
LGTM, merged. Thanks @person142.

Note that there's also non-float constants; integer flags like `np.CLIP`.

PR #53, opened 2020-04-16
ENH: add runtests option to find untyped module attributes

person142:
Now you can run something like

```
./runtests.py --find-missing numpy
```

to get a list of everything that still needs annotations added.

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -1,4 +1,7 @@
 #!/usr/bin/env python
+import argparse
+import ast
+import importlib
 import os
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


shoyer:
Nice! Any chance you could save the output in an issue somewhere for users to reference?

person142:
> Any chance you could save the output in an issue somewhere for users to reference?

The big list: https://github.com/numpy/numpy-stubs/issues/54

person142:
Anybody have any objections here? Otherwise I might merge in a bit so that updating the blacklist/list in the issue is easier.

shoyer:
Thanks!

person142:
Thanks @shoyer.

Issue #54, opened 2020-04-16
Big list of functions in the top-level namespace missing types


person142:
This is a big list of functions in the main `numpy` namespace that are missing types. You can regenerate an updated version of it with a

```
./runtests.py --find-missing numpy
```

If you are interested in contributing, adding stubs for one of these functions is a good place to start!

- [ ] AxisError
- [ ] ComplexWarning
- [ ] DataSource
- [ ] False_
- [ ] MachAr
- [ ] ModuleDeprecationWarning
- [ ] RankWarning
- [ ] ScalarType
- [ ] Tester
- [ ] TooHardError
- [ ] True_
- [ ] UFUNC_PYVALS_NAME
- [ ] VisibleDeprecationWarning
- [ ] abs
- [ ] absolute
- [ ] absolute_import
- [ ] add
- [ ] add_docstring
- [ ] add_newdoc
- [ ] add_newdoc_ufunc
- [ ] alen
- [ ] all
- [ ] alltrue
- [ ] amax
- [ ] amin
- [ ] angle
- [ ] any
- [ ] append
- [ ] apply_along_axis
- [ ] apply_over_axes
- [ ] arange
- [ ] arccos
- [ ] arccosh
- [ ] arcsin
- [ ] arcsinh
- [ ] arctan
- [ ] arctan2
- [ ] arctanh
- [ ] argmax
- [ ] argmin
- [ ] argpartition
- [ ] argsort
- [ ] around
- [ ] array2string
- [ ] array_repr
- [ ] array_split
- [ ] array_str
- [ ] asanyarray
- [ ] asarray
- [ ] asarray_chkfinite
- [ ] ascontiguousarray
- [ ] asfarray
- [ ] asfortranarray
- [ ] asmatrix
- [ ] asscalar
- [ ] atleast_1d
- [ ] atleast_2d
- [ ] atleast_3d
- [ ] average
- [ ] bartlett
- [ ] bincount
- [ ] bitwise_and
- [ ] bitwise_not
- [ ] bitwise_or
- [ ] bitwise_xor
- [ ] blackman
- [ ] block
- [ ] bmat
- [ ] bool
- [ ] bool8
- [ ] broadcast
- [ ] broadcast_arrays
- [ ] broadcast_to
- [ ] busday_count
- [ ] busday_offset
- [ ] busdaycalendar
- [ ] byte
- [ ] byte_bounds
- [ ] bytes0
- [ ] c_
- [ ] can_cast
- [ ] cast
[...]

PR #57, opened 2020-04-18
Added Warning and Exception annotations

BvB93:
See #54.

Added annotations and tests for:
* ~~``np.warnings``~~ (removed as of https://github.com/numpy/numpy-stubs/pull/57#discussion_r410731135)
* ``np.ModuleDeprecationWarning``
* ``np.VisibleDeprecationWarning``
* ``np.ComplexWarning``
* ``np.RankWarning``
* ``np.TooHardError``
* ``np.AxisError``


numpy-stubs/__init__.pyi
```diff
@@ -778,3 +779,17 @@ trunc: ufunc
 
 # TODO(shoyer): remove when the full numpy namespace is defined
 def __getattr__(name: str) -> Any: ...
+
+# Warnings
+warnings = _warnings
```
person142, line 784: Har, I think we should leave out warnings here and declare that any code using `np.warnings` is officially broken.

numpy-stubs/__init__.pyi
```diff
@@ -778,3 +779,17 @@ trunc: ufunc
...
+class ModuleDeprecationWarning(DeprecationWarning): ...
+class VisibleDeprecationWarning(UserWarning): ...
+class ComplexWarning(RuntimeWarning): ...
+class RankWarning(UserWarning): ...
+
+# Errors
+class TooHardError(RuntimeError): ...
+class AxisError(ValueError, IndexError):
+    # Note that all parameters for __init__ (except self) can technically be Any.
+    # So it's a question here whether or not we should allow what's possible or what's intended
```
person142, line 794: The way it is done here seems right to me.

tests/pass/warnings_and_errors.py
```diff
@@ -0,0 +1,18 @@
...
+ComplexWarning = np.ComplexWarning  # type: Type[RuntimeWarning]
+RankWarning = np.RankWarning  # type: Type[UserWarning]
+
+TooHardError = np.TooHardError  # type: Type[RuntimeError]
+AxisError1 = np.AxisError  # type: Type[ValueError]
+AxisError2 = np.AxisError  # type: Type[IndexError]
+
+np.AxisError(1)
+np.AxisError(1, ndim=2)
+np.AxisError(1, ndim=2, msg_prefix='error')
```
person142, line 18: Maybe add some cases in `tests/fail` for the constructor.

tests/pass/warnings_and_errors.py
```diff
@@ -0,0 +1,18 @@
+import types
+from typing import Type
+
+import numpy as np
+
+warnings = np.warnings  # type: types.ModuleType
+ModuleDeprecationWarning = np.ModuleDeprecationWarning  # type: Type[DeprecationWarning]
```
person142, line 7: I would move these to `test/reveal/warnings_and_errors.py` and do something like

```
reveal_type(ModuleDeprecationWarning())  # E: numpy.DeprecationWarning
...
```

numpy-stubs/__init__.pyi
person142: (Which means we can delete the comment I think.)

BvB93:
Thanks for feedback! I've addressed your points in the latest push.

person142:
Great, thanks @BvB93.

PR #58, opened 2020-04-18
Add Python 3.8 to the CI matrix

person142:
This also required bumping the flake8 version.

While experimenting with plugins in https://github.com/numpy/numpy-stubs/pull/56 I was getting CI failures on 3.7 but not 3.6, so it seemed prudent to expand our CI matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,12 @@
 language: python
 matrix:
   include:
+    - python: 3.8
+      dist: xenial
     - python: 3.7
       dist: xenial
-      sudo: true
     - python: 3.6
+      dist: xenial
 
 notifications:
   email: false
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,10 +1,12 @@
 language: python
 matrix:
   include:
+    - python: 3.8
+      dist: xenial
```
person142, line 5: Xenial is currently the default, but pinning it since it won't be the default forever.

.travis.yml
```diff
@@ -1,10 +1,12 @@
...
     - python: 3.7
       dist: xenial
-      sudo: true
```
person142, line 6: `sudo: true` is the default.

rgommers:

PR #62, opened 2020-04-19
MAINT: move __getattr__ to the top of __init__.pyi

person142:
We have a

```
def __getattr__(name: str) -> Any: ...
```

since not everything is typed yet. It was originally at the bottom of
the `__init__.pyi` file, but has since ended up not at the end because
it's hard to remember to put your code above it. Since it makes sense
to have it somewhere more noticeable than the middle of the file, put
it at the top instead.


rgommers:
Makes sense, thanks, merged

PR #63, opened 2020-04-20
Added typing_extensions as a dependency for Python < 3.8

BvB93:
* Added [``typing_extensions``](https://pypi.org/project/typing-extensions/) as a dependency for Python < 3.8.
* Ran black on setup.py.


person142:
Thanks @BvB93.

PR #64, opened 2020-04-21
MAINT: run black on all files in CI

person142:
In https://github.com/numpy/numpy-stubs/pull/63 black was run on `setup.py`. Follow up by running black on all files in CI. Do an inital pass over the entire codebase so that the CI check will pass.

I'm usually opposed to things like this, but it seems like we're in an early enough stage that it should be ok.

### DIFF
--- a/tests/fail/warnings_and_errors.py
+++ b/tests/fail/warnings_and_errors.py
@@ -2,4 +2,6 @@ import numpy as np
 
 np.AxisError(1.0)  # E: Argument 1 to "AxisError" has incompatible type
 np.AxisError(1, ndim=2.0)  # E: Argument "ndim" to "AxisError" has incompatible type
-np.AxisError(2, msg_prefix=404)  # E: Argument "msg_prefix" to "AxisError" has incompatible type
+np.AxisError(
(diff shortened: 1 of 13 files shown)


tests/fail/warnings_and_errors.py
```diff
@@ -2,4 +2,6 @@
 
 np.AxisError(1.0)  # E: Argument 1 to "AxisError" has incompatible type
 np.AxisError(1, ndim=2.0)  # E: Argument "ndim" to "AxisError" has incompatible type
-np.AxisError(2, msg_prefix=404)  # E: Argument "msg_prefix" to "AxisError" has incompatible type
+np.AxisError(
+    2, msg_prefix=404  # E: Argument "msg_prefix" to "AxisError" has incompatible type
```
person142, line 6: This is an interesting one where you do have to modify the output black gives you a bit because the exact line number matters. I don't think this should be a huge inconvenience going forward though.

rgommers:
This is fine with me, agreed that it's still early enough

person142:
Ok, let's merge it then before it gets stale.

Issue #37, opened 2019-11-24
What to do with array_like variables


johanvergeer:
I'm currently working on `core.numeric` and I'm running into an interesting issue.

A lot of methods have an `array_like` input parameter and an `ndarray` as the output parameter.

For now I created a `TypeVar` like the following:

```python
_ArrayLike = TypeVar("_ArrayLike", str, int, float, bool, object, ByteString, Iterable, Container, ndarray)
```

As an example I will use the `zeros_like` function

```python
def zeros_like(
        a: _ArrayLike,
        dtype: Optional[dtype] = None,
        order: str = 'K',
        subok: bool = True,
        shape: Optional[Union[int, Sequence[int]]] = None) -> ndarray[T]: ...
```

This function returns an instance of `ndarray` containing the same type as `a`, or it can be overridden by `dtype`. 

This brought me to the following conclusions (please correct me if I'm wrong)

1. `a` can be either 
   - `_ArrayLike` when `a` is a scalar or object
   - or `_ArrayLike[T]` when `a` is a collection
1. The return value type can be 
   - `ndarray[_ArrayLike]` when `a` is a scalar or object and `dtype` is `None`
   - `ndarray[T]` when `a` is `_ArrayLike[T]` and `dtype` is `None`
   - `ndarray[dtype]` when `dtype` is not `None`

I'm wondering what the type annotations of `a` and the return value should look like.

shoyer:
I don't think we've turned `ndarray` in a generic yet, which means the return value should probably just be `ndarray` for now?

rgommers:
Would it make sense to type `array_like` as a union of ndarray, sequence, scalar, and any/arraylike, rather than `_ArrayLike = TypeVar("_ArrayLike")`?

shoyer:
> Would it make sense to type `array_like` as a union of ndarray, sequence, scalar, and any/arraylike, rather than `_ArrayLike = TypeVar("_ArrayLike")`?

That seems like a reasonable start, though given the currently state of Numpy's aggressive coercion, arguably `Any` would be most correct. On the other hand, that would also nearly be useless for type checking, so we probably need to make some practical trade-off.

person142:
Looping back around to this, in a non-generic ndarray world, I think Ralf's suggestion of

> ndarray, sequence, scalar

(I took out `any`) is the right practical trade-off. I think that outside that world we get into the territory of "things that might surprise people", e.g.

```
>>> np.array((i for i in range(10))
```

[...]

Issue #13, opened 2018-03-06
Exporting type aliases for DtypeLike, ShapeLike and ArrayLike


shoyer:
These would indicate anything NumPy can coerce into the respective type:
- `DtypeLike` can be coerced into/with `np.dtype`
- `ShapeLike` can be coerced into `Tuple[int, ...]`
- `ArrayLike` can be coerced into `np.ndarray` with `np.array`

See https://github.com/numpy/numpy_stubs/pull/10 for (currently private) aliases `_DtypeLike` and `_ShapeLike`.

It would be nice to export these publicly for the benefit of downstream users.

The main complication is that this would likely also require adding these aliases to NumPy itself. Potentially, this could be unnecessary with `from __future__ import annotations` in Python 3.7 ([PEP 564](https://www.python.org/dev/peps/pep-0563/)).

JelleZijlstra:
Another option is for this repo to define a separate package called `numpy_typing` or similar that includes a `.py` defining these, so that users could do `from numpy_typing import DtypeLike`. With PEP 561 implemented, mypy should be able to pick up the definitions of the aliases from `numpy_typing`.

shoyer:
@JelleZijlstra That's a great alternative! Yes, we should definitely try the separate `numpy_typing` package to avoid the need to make upstream changes in NumPy yet.

flying-sheep:
numpy-mypy has those: https://github.com/machinalis/mypy-data/tree/master/numpy-mypy

person142:
This can be done without adding a separate package and without `from __future__ import annotations` (though that is certainly the neatest solution); the annotation just needs to be `'np.ShapeLike'` (e.g. a string).

person142:
I think we resolved all the issues here in https://github.com/numpy/numpy-stubs/pull/66.

PR #66, opened 2020-04-21
ENH: get more specific about _ArrayLike, make it public

person142:
Closes https://github.com/numpy/numpy-stubs/issues/37.

Add tests to check various examples.

### DIFF
--- a/numpy-stubs/__init__.pyi
+++ b/numpy-stubs/__init__.pyi
@@ -3,6 +3,8 @@ import sys
 import datetime as dt
 
 from numpy.core._internal import _ctypes
+from numpy.typing import ArrayLike, DtypeLike, _Shape, _ShapeLike
+
 from typing import (
     Any,
     ByteString,
@@ -36,69 +38,18 @@ else:
     from typing import SupportsBytes
 
 if sys.version_info >= (3, 8):
-    from typing import Literal
+    from typing import Literal, Protocol
 else:
-    from typing_extensions import Literal
+    from typing_extensions import Literal, Protocol
 
 # TODO: remove when the full numpy namespace is defined
 def __getattr__(name: str) -> Any: ...
 
-_Shape = Tuple[int, ...]
-
-# Anything that can be coerced to a shape tuple
-_ShapeLike = Union[int, Sequence[int]]
-
-_DtypeLikeNested = Any  # TODO: wait for support for recursive types
-
-# Anything that can be coerced into numpy.dtype.
-# Reference: https://docs.scipy.org/doc/numpy/reference/arrays.dtypes.html
-_DtypeLike = Union[
-    dtype,
-    # default data type (float64)
-    None,
-    # array-scalar types and generic types
-    type,  # TODO: enumerate these when we add type hints for numpy scalars
-    # TODO: add a protocol for anything with a dtype attribute
-    # character codes, type strings or comma-separated fields, e.g., 'float64'
-    str,
-    # (flexible_dtype, itemsize)
-    Tuple[_DtypeLikeNested, int],
-    # (fixed_dtype, shape)
-    Tuple[_DtypeLikeNested, _ShapeLike],
-    # [(field_name, field_dtype, field_shape), ...]
-    #
-    # The type here is quite broad because NumPy accepts quite a wide
-    # range of inputs inside the list; see the tests for some
-    # examples.
-    List[Any],
-    # {'names': ..., 'formats': ..., 'offsets': ..., 'titles': ...,
-    #  'itemsize': ...}
-    # TODO: use TypedDict when/if it's officially supported
-    Dict[
-        str,
-        Union[
-            Sequence[str],  # names
-            Sequence[_DtypeLikeNested],  # formats
-            Sequence[int],  # offsets
-            Sequence[Union[bytes, Text, None]],  # titles
-            int,  # itemsize
-        ],
-    ],
-    # {'field1': ..., 'field2': ..., ...}
-    Dict[str, Tuple[_DtypeLikeNested, int]],
-    # (base_dtype, new_dtype)
-    Tuple[_DtypeLikeNested, _DtypeLikeNested],
-]
-
 _NdArraySubClass = TypeVar("_NdArraySubClass", bound=ndarray)
-
-_ArrayLike = TypeVar("_ArrayLike")
 
 class dtype:
     names: Optional[Tuple[str, ...]]
-    def __init__(
-        self, obj: _DtypeLike, align: bool = ..., copy: bool = ...
-    ) -> None: ...
+    def __init__(self, obj: DtypeLike, align: bool = ..., copy: bool = ...) -> None: ...
     @property
     def alignment(self) -> int: ...
     @property
@@ -217,6 +168,7 @@ class _ArrayOrScalarCommon(
     def shape(self) -> _Shape: ...
     @property
     def strides(self) -> _Shape: ...
+    def __array__(self, __dtype: DtypeLike = ...) -> ndarray: ...
     def __int__(self) -> int: ...
     def __float__(self) -> float: ...
     def __complex__(self) -> complex: ...
@@ -299,7 +251,7 @@ class ndarray(_ArrayOrScalarCommon, Iterable, Sized, Container):
     def __new__(
         cls,
         shape: Sequence[int],
-        dtype: Union[_DtypeLike, str] = ...,
+        dtype: DtypeLike = ...,
         buffer: _BufferType = ...,
         offset: int = ...,
         strides: _ShapeLike = ...,
@@ -338,7 +290,7 @@ class ndarray(_ArrayOrScalarCommon, Iterable, Sized, Container):
     def dumps(self) -> bytes: ...
     def astype(
         self,
-        dtype: _DtypeLike,
+        dtype: DtypeLike,
         order: str = ...,
         casting: str = ...,
         subok: bool = ...,
@@ -349,14 +301,14 @@ class ndarray(_ArrayOrScalarCommon, Iterable, Sized, Container):
     @overload
     def view(self, dtype: Type[_NdArraySubClass]) -> _NdArraySubClass: ...
     @overload
-    def view(self, dtype: _DtypeLike = ...) -> ndarray: ...
+    def view(self, dtype: DtypeLike = ...) -> ndarray: ...
     @overload
     def view(
-        self, dtype: _DtypeLike, type: Type[_NdArraySubClass]
+        self, dtype: DtypeLike, type: Type[_NdArraySubClass]
     ) -> _NdArraySubClass: ...
     @overload
     def view(self, *, type: Type[_NdArraySubClass]) -> _NdArraySubClass: ...
-    def getfield(self, dtype: Union[_DtypeLike, str], offset: int = ...) -> ndarray: ...
+    def getfield(self, dtype: DtypeLike, offset: int = ...) -> ndarray: ...
     def setflags(
         self, write: bool = ..., align: bool = ..., uic: bool = ...
     ) -> None: ...
@@ -501,26 +453,26 @@ class str_(character): ...
 
 def array(
     object: object,
-    dtype: _DtypeLike = ...,
+    dtype: DtypeLike = ...,
     copy: bool = ...,
     subok: bool = ...,
     ndmin: int = ...,
 ) -> ndarray: ...
 def zeros(
-    shape: _ShapeLike, dtype: _DtypeLike = ..., order: Optional[str] = ...
+    shape: _ShapeLike, dtype: DtypeLike = ..., order: Optional[str] = ...
 ) -> ndarray: ...
 def ones(
-    shape: _ShapeLike, dtype: _DtypeLike = ..., order: Optional[str] = ...
+    shape: _ShapeLike, dtype: DtypeLike = ..., order: Optional[str] = ...
 ) -> ndarray: ...
 def zeros_like(
-    a: _ArrayLike,
+    a: ArrayLike,
     dtype: Optional[dtype] = ...,
     order: str = ...,
     subok: bool = ...,
     shape: Optional[Union[int, Sequence[int]]] = ...,
 ) -> ndarray: ...
 def ones_like(
-    a: _ArrayLike,
+    a: ArrayLike,
     dtype: Optional[dtype] = ...,
     order: str = ...,
     subok: bool = ...,
@@ -530,7 +482,7 @@ def full(
     shape: _ShapeLike, fill_value: Any, dtype: Optional[dtype] = ..., order: str = ...
 ) -> ndarray: ...
 def full_like(
-    a: _ArrayLike,
+    a: ArrayLike,
     fill_value: Any,
     dtype: Optional[dtype] = ...,
     order: str = ...,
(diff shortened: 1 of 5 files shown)


person142:
As something of a data point, MyPy passes on SciPy when checked against this branch. Now the types in SciPy are still _very_ rough, so take that with a grain of salt, but perhaps it means something.

rgommers:
What about other array-like objects? Are things with both `__array__` and `__len__` passing because of `Sequence`, or is that not tested? Or is that what the Protocol discussion was about?

numpy-stubs/__init__.pyi
```diff
@@ -87,7 +87,7 @@ _DtypeLike = Union[
 
 _NdArraySubClass = TypeVar("_NdArraySubClass", bound=ndarray)
 
-_ArrayLike = TypeVar("_ArrayLike")
+ArrayLike = Union[int, float, complex, generic, ndarray, Sequence]
```
BvB93, line 90: Might it be an idea to, respectively, replace ``int``, ``float`` and ``complex`` with the ``SupportsInt``, ``SupportsFloat`` and ``SupportsComplex`` protocols?

What about ``str``, ``bytes`` and ``dt.datetime`` (considering they all have their corresponding generic)?

Furthermore, I'd presonally be in favor to replace ``ndarray`` with a custom ``_SupportsArray`` protocol along the lines of:
``` python
class _SupportsArray(Protocol):
    def __array__(self, dtype: _DtypeLike = ...) -> ndarray: ...
```

person142:
@rgommers @BvB93 thanks for the feedback-I added a `_SupportsArray` protocol in https://github.com/numpy/numpy-stubs/pull/66/commits/4f654e538115c8c741dcd96b0b64d5ab8d8ef80e; it is quite a nice unification.

numpy-stubs/__init__.pyi
person142: > Might it be an idea to, respectively, replace int, float and complex with the SupportsInt, SupportsFloat and SupportsComplex protocols?

I think the protocols are a little too general; e.g.

```
>>> class A:
...     def __int__(self):
...         return 1
...
>>> int(A())
1
>>> np.array(A())
array(<__main__.A object at 0x10e61a290>, dtype=object)
```

> What about str, bytes, bool, dt.datetime dt.date and dt.timedelta (considering they all have their corresponding generic)?

`str` and `bytes` are both covered by `Sequence`. I added bool (thanks for catching that!). The `dt.*` I think also ends up giving unexpected results:

```
>>> np.array(datetime.timedelta(days=1))
array(datetime.timedelta(days=1), dtype=object)
```

> Furthermore, I'd presonally be in favor of replacing ndarray with a custom _SupportsArray protocol along the lines of:

Yes, that is an excellent idea, switched to that.

person142:
It seems that the principle we're roughly going for here is "don't allow stuff that will produce object arrays", though we still leave an escape hatch a la

https://github.com/numpy/numpy-stubs/commit/4f654e538115c8c741dcd96b0b64d5ab8d8ef80e#diff-98e9e1660b68614cffb8585ea52a0bdcR31.

numpy-stubs/__init__.pyi
BvB93: > I think the protocols are a little too general; e.g.

Good catch, I actually wasn't aware that, _e.g._, a ``SupportsInt`` member wouldn't produce an integer array.

> `str` and `bytes` are both covered by `Sequence`

Ah right, that's true.

BvB93:
> 
> 
> @rgommers @BvB93 thanks for the feedback-I added a `_SupportsArray` protocol in [4f654e5](https://github.com/numpy/numpy-stubs/commit/4f654e538115c8c741dcd96b0b64d5ab8d8ef80e); it is quite a nice unification.

Btw, doesn't ``__array__()`` also have the ``dtype`` argument?

person142:
> Btw, doesn't __array__() also have the dtype argument?

Hm, it appears to be a little wonky:

```python
>>> np.float64(1).__array__()
array(1.)
>>> np.float64(1).__array__(np.complex128)
array(1.+0.j)
>>> np.float64(1).__array__(dtype=np.complex128)
TypeError: __array__() takes no keyword arguments
>>> class A:
...     def __array__(self, dtype):
...         return np.array([1, 2, 3], dtype=dtype)
...
>>> np.array(A())
TypeError: __array__() missing 1 required positional argument: 'dtype'
>>> np.array(A(), dtype=np.float64)
array([1., 2., 3.])
array([1., 2., 3.])
>>> class A:
...     def __array__(self, dtype=None):
...         return np.array([1, 2, 3], dtype=dtype)
...
>>> np.array(A())
array([1, 2, 3])
>>> np.array(A(), dtype=np.float64)
array([1., 2., 3.])
```

person142:
That's unfortunate... it looks like the `__array__` methods on scalars are inherently different from the user ones since they force positional arguments, whereas only positional arguments in the user ones don't work.

person142:
Ok, I made the protocol

```python
class _SupportsArray(Protocol):
    def __array__(
        self, dtype: Optional[_DTypeLike] = ...
    ) -> Union[Sequence, ndarray]: ...
```

This unfortunately means that `ndarray` and `generic` no longer satisfy it! To work around that I added them back to the big union. This is kind of gross, but I think it's the best we can do given the above examples. WDYT?

BvB93:
> This unfortunately means that ndarray and generic no longer satisfy it! To work around that I added them back to the big union. This is kind of gross, but I think it's the best we can do given the above examples. WDYT?

I'd suggest adding another overload to the ``_SupportsArray`` protocol instead such that it works with in situations where ``dtype`` is [positional-only](https://www.python.org/dev/peps/pep-0484/#positional-only-arguments) or keyword-or-positional:

``` python
# first overload: dtype is optional and positional-only
# Second overload: dtype is optional and can be a positional or keyword argument
class _SupportsArray(Protocol):
    @overload
    def __array__(self, __dtype: _DtypeLike = ...) -> ndarray: ...
    @overload
    def __array__(self, dtype: _DtypeLike = ...) -> ndarray: ...


class A():
    def __array__(self, dtype: _DtypeLike = None) -> ndarray:
        return np.array([1, 2, 3], dtype=dtype)

class B():
    def __array__(self, __dtype: _DtypeLike = None) -> ndarray:
        return np.array([1, 2, 3], dtype=__dtype)

a = A()
a.__array__()
a.__array__(float)
a.__array__(dtype=float)

b = B()
b.__array__()
b.__array__(float)
b.__array__(dtype=float)  # E: Unexpected keyword argument
b.__array__(__dtype=float)  # E: Unexpected keyword argument
```

By the way, why is the returned type annotated as ``Union[Sequence, ndarray]`` in your example? Doesn't it always just return ``ndarray``?

person142:
> I'd suggest adding another overload to the _SupportsArray protocol instead, such that it works with in situations where dtype is positional-only or keyword-or-positional

The signatures are incompatible, so that’s going to violate Liskov I think.

> By the way, why is the returned type annotated as Union[Sequence, ndarray] in your example?

I was going by the docs here:

https://docs.scipy.org/doc/numpy/reference/generated/numpy.array.html

which say ndarray or nested sequence.

BvB93:
> The signatures are incompatible, so that’s going to violate Liskov I think. (But I’ll give it a try.)

It's not quite clear to me where this incompatible is located.
``__array__`` ([ref](https://docs.scipy.org/doc/numpy/reference/arrays.classes.html#numpy.class.__array__)) takes ``self`` and (optionally) ``dtype`` as argument. The newly overload ``_SupportsArray`` can now deal with where ``dtype`` is either keyword-only or positional-or-keyword. 

> I was going by the docs here:

I believe what they're talking about in the docs is an object which either:
a. Has an ``__array__()`` method that returns an array.
b. Is a (nested) sequence.

So ``Union[_SupportsArray, Sequence[Any]]`` in other words, rather than ``__array__`` itself (potentially) returning a Sequence.

person142:
> It's not quite clear to me where this incompatible is located.

> I believe what they're talking about in the docs is an object which either:

Yup, right on both counts. Updated; hopefully we're good to go now.

Note that I had to make `_DtypeLike` public too so that someone could use it in the annotations for their `__array__` method, but that's ok because it was desired anyway (https://github.com/numpy/numpy-stubs/issues/13).

shoyer:
This looks really nice!

My own serious concern here is about adding public type aliases. These do seem quite useful, but what would this imply if/when we move type annotations into NumPy properly? If `np.ArrayLike` needs to be valid, does that imply that if we move annotations into NumPy we need to expose `ArrayLike` in the public API of NumPy?

person142:
> If np.ArrayLike needs to be valid, does that imply that if we move annotations into NumPy we need to expose ArrayLike in the public API of NumPy?

Thankfully it doesn't-this is the same sort of fudging that happens with things like `Queue[T]` in the stdlib (where the real class isn't actually generic). But it does mean that people using annotations need to take one of these options:

```python
import numpy as np

x: "np.ArrayLike"  # Use strings
```

```python
from __future__ import annotations

import numpy as np

x: np.ArrayLike  # Now it's treated as a string anyway
```

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from numpy import ArrayLike
else:
    ArrayLike = None  # Or whatever

x: ArrayLike
```
```

**Comment by person142** [comment]
Although maybe I am interpreting your question in the wrong way. I think a backdrop to my answer above is an assumption that even when we move the types into NumPy, they will remain stubs instead of being inlined into the code.

I suspect that this is the right course of action because we've seen that the types require a fair bit of "fudging"; i.e. we aren't trying to represent the full NumPy API but instead some typeable subset of it. I think that if the types are inlined then we lose the ability to do that fudging as well.

I'd contrast this to e.g. SciPy where we are inlining the types, mostly because there isn't much odd there (except for needing a bunch of stubs for extension modules).

**Comment by ethanhs** [comment]
FWIW, I've heard that at the language summit they decided `__future__.annotations` will become default in 3.10 I believe.

**Comment by shoyer** [comment]
I appreciate that we could need np.ArrayLike in strings or type annotations, but I suspect that it could still lead to some user confusion to not define them at runtime, too, e.g., to cover use cases like type aliases. The user experience is independent of whether we choose to use stubs or annotations inside NumPy — though I expect we’ll probably end up with some of both.

NumPy-stubs is certainly still experimental, so you definitely have my blessing to go ahead for now. But I do think it could be worth sounding out the broader NumPy community on the appetite for adding a selective handful of type protocols into NumPy proper. We might get some useful feedback. For example, should the protocol be called `ArrayLike` or `array_like` as currently appears in many NumPy docstrings?

**Comment by rgommers** [comment]
If it helps making `ArrayLike` et al. public, I don't see a big issue with that; would probably prefer to write it as 
```
from numpy.types import ArrayLike

x: ArrayLike
```
to keep it out of the main namespace. In my (limited) experience, it's helpful for things to exist at runtime.

**Comment by person142** [comment]
> from numpy.types import ArrayLike

Keeping it out of the main namespace seems good, though maybe it should be `numpy.typing` instead of `numpy.types` to match the stdlib module?

**Comment by person142** [comment]
Re

> But I do think it could be worth sounding out the broader NumPy community on the appetite for adding a selective handful of type protocols into NumPy proper.

I'll send out something to the mailing list.

**Comment by rgommers** [comment]
> > from numpy.types import ArrayLike
> 
> Keeping it out of the main namespace seems good, though maybe it should be `numpy.typing` instead of `numpy.types` to match the stdlib module?

No, I avoided that name on purpose, naming something the same as a stdlib module is usually a bad idea (e.g. `scipy.io` was a pretty bad choice, that's why it's usually import as `spio`).

**Comment by BvB93** [comment]
What about something along the lines of ``numpy.annotations``?

**Comment by eric-wieser** [comment]
Althou

> > > from numpy.types import ArrayLike
> > 
> > 
> > Keeping it out of the main namespace seems good, though maybe it should be `numpy.typing` instead of `numpy.types` to match the stdlib module?
> 
> No, I avoided that name on purpose, naming something the same as a stdlib module is usually a bad idea


As a counterargument, `numpy.distutils` clashes with `distutils`, but we used it anyway, presumably because the similarity was worth emphasizing.

I'd imagine if we use `numpy.typing`, people would be happy to use something like
```
def get_arr() -> typing.Tuple[np.typing.ArrayLike, int]: ...
```

**Comment by person142** [comment]
Ok, discussion on the mailing list:

http://numpy-discussion.10968.n7.nabble.com/Feelings-about-type-aliases-in-NumPy-td48059.html

seems to be dying down. Takeaways so far:

- No (open) objections to adding types to NumPy itself in the future
- Most people preferred not doing it in the top-level namespace; though @shoyer was a notable objector
- Little agreement on what to name the module

So, as is often the case, seems like there's rough consensus except on what to name the darned thing.

File: numpy-stubs/__init__.pyi [hunk 2]
```diff
@@ -92,13 +92,17 @@ _DtypeLike = Union[
 
 _NdArraySubClass = TypeVar("_NdArraySubClass", bound=ndarray)
 
-_ArrayLike = TypeVar("_ArrayLike")
+class _SupportsArray(Protocol):
+    @overload
+    def __array__(self, __dtype: DtypeLike = ...) -> ndarray: ...
+    @overload
+    def __array__(self, dtype: Optional[DtypeLike] = ...) -> ndarray: ...
```
BvB93, line 99: Is the ``Optional`` in ``Optional[DtypeLike]`` not redundant here?
Considering ``None`` is already included the union of ``DtypeLike``.

numpy-stubs/__init__.pyi
```diff
@@ -217,6 +221,7 @@ class _ArrayOrScalarCommon(
     def shape(self) -> _Shape: ...
     @property
     def strides(self) -> _Shape: ...
+    def __array__(self, __dtype: Optional[DtypeLike] = ...) -> ndarray: ...
```
BvB93, line 224: See comment above.

numpy-stubs/__init__.pyi
```diff
@@ -349,14 +354,14 @@ class ndarray(_ArrayOrScalarCommon, Iterable, Sized, Container):
...
+    def view(self, dtype: DtypeLike = ...) -> ndarray: ...
     @overload
     def view(
-        self, dtype: _DtypeLike, type: Type[_NdArraySubClass]
+        self, dtype: DtypeLike, type: Type[_NdArraySubClass]
     ) -> _NdArraySubClass: ...
     @overload
     def view(self, *, type: Type[_NdArraySubClass]) -> _NdArraySubClass: ...
-    def getfield(self, dtype: Union[_DtypeLike, str], offset: int = ...) -> ndarray: ...
+    def getfield(self, dtype: Union[DtypeLike, str], offset: int = ...) -> ndarray: ...
```
eric-wieser, line 364: Can this be simplified?
```suggestion
    def getfield(self, dtype: DtypeLike, offset: int = ...) -> ndarray: ...
```

TomAugspurger:
cc @simonjayhawkins if you see any issues with how this ArrayLike will interact with pandas' typing.

person142:
I don't want us to lose momentum here, so: I find @eric-wieser's comment

> Almost all the usage of type annotations I've seen in the wild has erred on the side of keeping the annotations as short as possible

to match with my own experience as well-typing in Python is fairly verbose, so short forms like

```
import typing as t
```

or

```
from typing import ...
```

seem to be the norm. For that reason I propose that we move ahead with putting things in `numpy.typing`. For now it will be in the stubs only, and when we merge the stubs into NumPy itself we can make it available at runtime.

Are people ok with that, or shall be continue to discuss?

shoyer:
+1 for numpy.typing, but please send this to the email list, which is the
place of record for design decisions

rgommers:
Sounds fine to me, thanks for keeping this moving.

person142:
Ok, mailing list has been notified, PR has been rebased, review comments have been addressed (I think), the types have been moved into `numpy.typing`, and the tests have been updated accordingly.

person142:
Any objections to moving forward? Since this conflicts with just about every other PR it would be nice to get it in to avoid more rebasing.

BvB93:
No complaints here from my side; feel free to continue.

shoyer:
Look good to me!

person142:
In it goes then. Thanks for reviewing everyone!

PR #59, opened 2020-04-19
WiP: Added type annotations for the ``np.core.fromnumeric`` module

BvB93:
See https://github.com/numpy/numpy-stubs/issues/54.

Introduced type hints for all (public) functions from the [``np.core.fromnumeric``](https://github.com/numpy/numpy/blob/master/numpy/core/fromnumeric.py) module.

~~Note that [``typing_extensions``](https://pypi.org/project/typing-extensions/) has also been added as a dependency for Python version prior to 3.8.~~ See https://github.com/numpy/numpy-stubs/pull/63.

Functions
----------
* ``take``
* ``reshape``
* ``choose``
* ``repeat``
* ``put``
* ``swapaxes``
* ``transpose``
* ``partition``
* ``argpartition``
* ``sort``
* ``argsort``
* ``argmax``
* ``argmin``
* ``searchsorted``
* ``resize``
* ``squeeze``
* ``diagonal``
* ``trace``
* ``ravel``
* ``nonzero``
* ``shape``
* ``compress``
* ``clip``
* ``sum``
* ``all``
* ``any``
* ``cumsum``
* ``ptp``
* ``amax``
* ``amin``
* ``alen``
* ``prod``
* ``cumprod``
* ``ndim``
* ``size``
* ``around``
* ``mean``
* ``std``
* ``var``
* ``round_``, ``product``, ``cumproduct``, ``sometrue`` and ``alltrue``

To do
------
- [x] Add reveal tests
- [ ] Add pass tests
- [ ] Add fail tests

PR #67, opened 2020-04-21
ENH: Add type annotations for the np.core.fromnumeric module: part 1/4

BvB93:
Continuation of https://github.com/numpy/numpy-stubs/pull/59.

Introduced type hints for the first 11 (public) functions from the [``np.core.fromnumeric``](https://github.com/numpy/numpy/blob/master/numpy/core/fromnumeric.py) module.

Functions
----------
* ``take``
* ``reshape``
* ``choose``
* ``repeat``
* ``put``
* ``swapaxes``
* ``transpose``
* ``partition``
* ``argpartition``
* ``sort``
* ``argsort``

### DIFF
--- a/numpy-stubs/__init__.pyi
+++ b/numpy-stubs/__init__.pyi
@@ -35,6 +35,11 @@ if sys.version_info[0] < 3:
 else:
     from typing import SupportsBytes
 
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
+
 # TODO: remove when the full numpy namespace is defined
 def __getattr__(name: str) -> Any: ...
 
@@ -792,3 +797,124 @@ class AxisError(ValueError, IndexError):
     def __init__(
         self, axis: int, ndim: Optional[int] = ..., msg_prefix: Optional[str] = ...
     ) -> None: ...
+
+# Functions from np.core.fromnumeric
+_Mode = Literal["raise", "wrap", "clip"]
+_Order = Literal["C", "F", "A"]
+_PartitionKind = Literal["introselect"]
+_SortKind = Literal["quicksort", "mergesort", "heapsort", "stable"]
+
+_Generic = TypeVar("_Generic", bound=generic)
+
+# Can't use the SupportsInt protocol here as the ndarray.__int__() method exists
+_ScalarBuiltin = Union[str, bytes, dt.date, dt.timedelta, bool, int, float, complex]
+_Scalar = Union[_ScalarBuiltin, generic]
+
+# An array-like object consisting of integers
+_ArrayLikeInt = Union[
+    int, integer, Sequence[Union[int, integer]], ndarray
+]  # TODO: ndarray[int]
+
+# An array-like object consisting of strings
+_ArrayLikeStr = Union[
+    str, str_, Sequence[Union[str, str_]], ndarray
+]  # TODO: ndarray[str]
+
+# The signature of take() follows a common theme with its overloads:
+# 1. A generic comes in; the same generic comes out
+# 2. A scalar comes in; a generic comes out
+# 3. An array-like object comes in; some keyword ensures that a generic comes out
+# 4. An array-like object comes in; an ndarray or generic comes out
+@overload
+def take(
+    a: _Generic,
+    indices: int,
+    axis: Optional[int] = ...,
+    out: Optional[ndarray] = ...,
+    mode: _Mode = ...,
+) -> _Generic: ...
+@overload
+def take(
+    a: _Scalar,
+    indices: int,
+    axis: Optional[int] = ...,
+    out: Optional[ndarray] = ...,
+    mode: _Mode = ...,
+) -> generic: ...
+@overload
+def take(
+    a: _ArrayLike,
+    indices: int,
+    axis: Optional[int] = ...,
+    out: Optional[ndarray] = ...,
+    mode: _Mode = ...,
+) -> generic: ...
+@overload
+def take(
+    a: _ArrayLike,
+    indices: _ArrayLikeInt,
+    axis: Optional[int] = ...,
+    out: Optional[ndarray] = ...,
+    mode: _Mode = ...,
+) -> Union[generic, ndarray]: ...
+def reshape(a: _ArrayLike, newshape: _ShapeLike, order: _Order = ...) -> ndarray: ...
+@overload
+def choose(
+    a: _Generic,
+    choices: Union[Sequence[_ArrayLike], ndarray],  # TODO: ndarray[_ArrayLike]
+    out: Optional[ndarray] = ...,
+    mode: _Mode = ...,
+) -> _Generic: ...
+@overload
+def choose(
+    a: _Scalar,
+    choices: Union[Sequence[_ArrayLike], ndarray],  # TODO: ndarray[_ArrayLike]
+    out: Optional[ndarray] = ...,
+    mode: _Mode = ...,
+) -> generic: ...
+@overload
+def choose(
+    a: _ArrayLike,
+    choices: Union[Sequence[_ArrayLike], ndarray],  # TODO: ndarray[_ArrayLike]
+    out: Optional[ndarray] = ...,
+    mode: _Mode = ...,
+) -> ndarray: ...
+def repeat(
+    a: _ArrayLike, repeats: _ArrayLikeInt, axis: Optional[int] = ...
+) -> ndarray: ...
+def put(a: ndarray, ind: _ArrayLikeInt, v: _ArrayLike, mode: _Mode = ...) -> None: ...
+def swapaxes(
+    a: Union[Sequence[_ArrayLike], ndarray],  # TODO: ndarray[_ArrayLike]
+    axis1: int,
+    axis2: int,
+) -> ndarray: ...
+def transpose(
+    a: _ArrayLike,
+    axes: Union[None, Sequence[int], ndarray] = ...,  # TODO: ndarray[int]
+) -> ndarray: ...
+def partition(
+    a: _ArrayLike,
+    kth: _ArrayLikeInt,
+    axis: Optional[int] = ...,
+    kind: _PartitionKind = ...,
+    order: Optional[_ArrayLikeStr] = ...,
+) -> ndarray: ...
+def argpartition(
+    a: _ArrayLike,
+    kth: _ArrayLikeInt,
+    axis: Optional[int] = ...,
+    kind: _PartitionKind = ...,
+    order: Optional[_ArrayLikeStr] = ...,
+) -> ndarray: ...
+def sort(
+    a: Union[Sequence[_ArrayLike], ndarray],
+    axis: Optional[int] = ...,
+    kind: Optional[_SortKind] = ...,
+    order: Optional[_ArrayLikeStr] = ...,
+) -> ndarray: ...
+def argsort(
+    a: Union[Sequence[_ArrayLike], ndarray],
+    axis: Optional[int] = ...,
+    kind: Optional[_SortKind] = ...,
+    order: Optional[_ArrayLikeStr] = ...,
+) -> ndarray: ...
--- a/numpy-stubs/__init__.pyi
+++ b/numpy-stubs/__init__.pyi
@@ -813,10 +813,7 @@ _Scalar = Union[_ScalarBuiltin, generic]
 _Int = Union[int, integer]
 _ArrayLikeIntNested = Any  # TODO: wait for support for recursive types
 _ArrayLikeInt = Union[
-    _Int,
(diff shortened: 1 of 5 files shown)


numpy-stubs/__init__.pyi
```diff
@@ -792,3 +797,124 @@ class AxisError(ValueError, IndexError):
...
+
+# Functions from np.core.fromnumeric
+_Mode = Literal["raise", "wrap", "clip"]
+_Order = Literal["C", "F", "A"]
+_PartitionKind = Literal["introselect"]
+_SortKind = Literal["quicksort", "mergesort", "heapsort", "stable"]
+
+_Generic = TypeVar("_Generic", bound=generic)
+
+# Can't use the SupportsInt protocol here as the ndarray.__int__() method exists
```
BvB93, line 809: It might be worth considering removing ``__int__()``, ``__float__()`` and ``__complex__()`` from ``ndarray`` for now; at least until there is proper support for array shapes (as only 0D arrays actually support the use of these methods). 

This would most certainly make it easier to identify (numerical) scalars.

BvB93:
It seems there are some minor differences between the formatting style of ``black`` ``18.9b0`` and the later ``19.10b0`` version, to the point were the latter will fail the tests (as ``18.9b0`` is pinned in test-requirements.txt)

numpy-stubs/__init__.pyi
BvB93: Scratch this idea, the use of Protocols in identifying scalars might not be so straightforward as I initially thought (see https://github.com/numpy/numpy-stubs/pull/66#discussion_r412235311).

numpy-stubs/__init__.pyi
```diff
@@ -792,3 +797,123 @@ class AxisError(ValueError, IndexError):
...
+    a: _ArrayLike,
+    indices: int,
+    axis: Optional[int] = ...,
+    out: Optional[ndarray] = ...,
+    mode: _Mode = ...,
+) -> generic: ...
+@overload
+def take(
+    a: _ArrayLike,
+    indices: _ArrayLikeInt,
```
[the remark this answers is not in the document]
BvB93: That's true, I forgot the ``Sequence[int]`` syntax does not support arbitrary levels of nesting (which is what we need here).
I don't have a real solution for this, surprisingly fundamental, problem besides adding ``Union``s up to some arbitrary level of nesting that we can consider "good enough".

numpy-stubs/__init__.pyi
```diff
@@ -792,3 +797,123 @@ class AxisError(ValueError, IndexError):
...
+def choose(
+    a: _ArrayLike,
+    choices: Union[Sequence[_ArrayLike], ndarray],  # TODO: ndarray[_ArrayLike]
+    out: Optional[ndarray] = ...,
+    mode: _Mode = ...,
+) -> ndarray: ...
+def repeat(
+    a: _ArrayLike, repeats: _ArrayLikeInt, axis: Optional[int] = ...
+) -> ndarray: ...
+def put(a: ndarray, ind: _ArrayLikeInt, v: _ArrayLike, mode: _Mode = ...) -> None: ...
```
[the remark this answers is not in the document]
BvB93: I have no clue, but the same issue applies as noted in the previous comment.

numpy-stubs/__init__.pyi
```diff
@@ -792,3 +797,123 @@ class AxisError(ValueError, IndexError):
...
         self, axis: int, ndim: Optional[int] = ..., msg_prefix: Optional[str] = ...
     ) -> None: ...
+
+# Functions from np.core.fromnumeric
+_Mode = Literal["raise", "wrap", "clip"]
+_Order = Literal["C", "F", "A"]
+_PartitionKind = Literal["introselect"]
+_SortKind = Literal["quicksort", "mergesort", "heapsort", "stable"]
+
+_Generic = TypeVar("_Generic", bound=generic)
```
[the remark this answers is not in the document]
BvB93: I feel this is a sensible change.

tests/fail/fromnumeric.py
```diff
@@ -0,0 +1,60 @@
+import numpy as np
```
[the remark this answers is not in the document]
BvB93: Currently it's named after the corresponding module in NumPy (``np.core.fromnumeric``).
I'm all ears if you have suggestions, but I feel that the current name is already pretty decent.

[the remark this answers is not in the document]
BvB93: Im any case, I've added a module-level docstring to the tests for now to clarify that the relevant functions are from ``np.core.fromnumeric`` (https://github.com/numpy/numpy-stubs/pull/67/commits/37cbdedd2a60bbb92c1eb3b52b65a4c6786058ec).

numpy-stubs/__init__.pyi
```diff
@@ -792,3 +797,131 @@ class AxisError(ValueError, IndexError):
...
+
+# Various annotations for scalars
+_ScalarGeneric = TypeVar("_ScalarGeneric", bound=generic)
+_ScalarBuiltin = Union[str, bytes, dt.date, dt.timedelta, bool, int, float, complex]
+_Scalar = Union[_ScalarBuiltin, generic]
+
+# An array-like object consisting of integers
+# TODO: If possible, figure out a better way to deal with nested sequences
+_Int = Union[int, integer]
+_ArrayLikeInt = Union[
```
BvB93, line 815: Continuation of https://github.com/numpy/numpy-stubs/pull/67#discussion_r412327691.
Not the most elegant solution, but it sort of does the job.

numpy-stubs/__init__.pyi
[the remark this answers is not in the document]
person142: So far we've just been doing `Sequence` (i.e. `Sequence[Any]`). I think the long-term fix is MyPy et. al. supporting recursive types; until then we're stuck with being imprecise.

It's not the best, but I think it's what we have right now. (Though I have also considered the 

> adding Unions up to some arbitrary level of nesting

It felt too gross to do, but maybe there's more discussion to be had on it in an issue.)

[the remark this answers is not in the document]
BvB93: > It felt too gross to do, but maybe there's more discussion to be had on it in an issue.)

Gross it most definetly is; I've changed it to ``Sequence[Any]`` for now (https://github.com/numpy/numpy-stubs/pull/67/commits/837eec640b4cca8616f316e0542f0cfd025ab4a5).

numpy-stubs/__init__.pyi
```diff
@@ -792,3 +797,124 @@ class AxisError(ValueError, IndexError):
...
+    mode: _Mode = ...,
+) -> _ScalarGeneric: ...
+@overload
+def take(
+    a: _Scalar,
+    indices: int,
+    axis: Optional[int] = ...,
+    out: Optional[ndarray] = ...,
+    mode: _Mode = ...,
+) -> generic: ...
```
person142, line 844: This isn't always true I don't think:

```python
>>> type(np.take(datetime.timedelta(days=1), 0))
<class 'datetime.timedelta'>
>>> type(np.take(datetime.date.today(), 0))
<class 'datetime.date'>
```

Works for the others though. There might be diminishing returns getting this to work until such a time as we have a templating system in place; then it can generate specific overloads like `bool -> np.bool_` etc.

numpy-stubs/__init__.pyi
```diff
@@ -792,3 +797,124 @@ class AxisError(ValueError, IndexError):
...
+) -> ndarray: ...
+def transpose(
+    a: _ArrayLike, axes: Union[None, Sequence[int], ndarray] = ...  # TODO: ndarray[int]
+) -> ndarray: ...
+def partition(
+    a: _ArrayLike,
+    kth: _ArrayLikeInt,
+    axis: Optional[int] = ...,
+    kind: _PartitionKind = ...,
+    order: Optional[_ArrayLikeStr] = ...,
```
person142, line 900: This appears to be an instance where we don't want `ndarray`:

```python
>>> x = np.zeros((10,), dtype=[('foo', np.float64), ('bar', np.float64)])
>>> np.partition(x, kth=0, order=['foo', 'bar'])
array([(0., 0.), (0., 0.), (0., 0.), (0., 0.), (0., 0.), (0., 0.),
       (0., 0.), (0., 0.), (0., 0.), (0., 0.)],
      dtype=[('foo', '<f8'), ('bar', '<f8')])
>>> np.partition(x, kth=0, order=('foo', 'bar'))
array([(0., 0.), (0., 0.), (0., 0.), (0., 0.), (0., 0.), (0., 0.),
       (0., 0.), (0., 0.), (0., 0.), (0., 0.)],
      dtype=[('foo', '<f8'), ('bar', '<f8')])
>>> np.partition(x, kth=0, order=np.array(['foo', 'bar']))
  File "<__array_function__ internals>", line 6, in partition
  File "/Users/Josh/Projects/numpy-stubs-dev/venv/lib/python3.7/site-packages/numpy/core/fromnumeric.py", line 746, in partition
    a.partition(kth, axis=axis, kind=kind, order=order)
  File "/Users/Josh/Projects/numpy-stubs-dev/venv/lib/python3.7/site-packages/numpy/core/_internal.py", line 383, in _newnames
    raise ValueError("unsupported order value: %s" % (order,))
ValueError: unsupported order value: ['foo' 'bar']
```

So I think here and below should be `Optional[Union[str, Sequence[str]]`.

numpy-stubs/__init__.pyi
```diff
@@ -792,3 +797,124 @@ class AxisError(ValueError, IndexError):
...
+    a: Union[Sequence[_ArrayLike], ndarray],
+    axis: Optional[int] = ...,
+    kind: Optional[_SortKind] = ...,
+    order: Optional[_ArrayLikeStr] = ...,
+) -> ndarray: ...
+def argsort(
+    a: Union[Sequence[_ArrayLike], ndarray],
+    axis: Optional[int] = ...,
+    kind: Optional[_SortKind] = ...,
+    order: Optional[_ArrayLikeStr] = ...,
```
person142, line 919: Similar comment as above:

```python
>>> np.argsort(x, order=['foo', 'bar'])
array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
>>> np.argsort(x, order=np.array(['foo', 'bar']))
  File "<__array_function__ internals>", line 6, in argsort
  File "/Users/Josh/Projects/numpy-stubs-dev/venv/lib/python3.7/site-packages/numpy/core/fromnumeric.py", line 1105, in argsort
    return _wrapfunc(a, 'argsort', axis=axis, kind=kind, order=order)
  File "/Users/Josh/Projects/numpy-stubs-dev/venv/lib/python3.7/site-packages/numpy/core/fromnumeric.py", line 61, in _wrapfunc
    return bound(*args, **kwds)
  File "/Users/Josh/Projects/numpy-stubs-dev/venv/lib/python3.7/site-packages/numpy/core/_internal.py", line 383, in _newnames
    raise ValueError("unsupported order value: %s" % (order,))
ValueError: unsupported order value: ['foo' 'bar']
```

numpy-stubs/__init__.pyi
```diff
@@ -792,3 +797,124 @@ class AxisError(ValueError, IndexError):
...
+    kth: _ArrayLikeInt,
+    axis: Optional[int] = ...,
+    kind: _PartitionKind = ...,
+    order: Optional[_ArrayLikeStr] = ...,
+) -> ndarray: ...
+def sort(
+    a: Union[Sequence[_ArrayLike], ndarray],
+    axis: Optional[int] = ...,
+    kind: Optional[_SortKind] = ...,
+    order: Optional[_ArrayLikeStr] = ...,
```
person142, line 913: Similar comment as above:

```python
>>> np.sort(x, order=np.array(['foo', 'bar']))
  File "<__array_function__ internals>", line 6, in sort
  File "/Users/Josh/Projects/numpy-stubs-dev/venv/lib/python3.7/site-packages/numpy/core/fromnumeric.py", line 989, in sort
    a.sort(axis=axis, kind=kind, order=order)
  File "/Users/Josh/Projects/numpy-stubs-dev/venv/lib/python3.7/site-packages/numpy/core/_internal.py", line 383, in _newnames
    raise ValueError("unsupported order value: %s" % (order,))
ValueError: unsupported order value: ['foo' 'bar']
```

numpy-stubs/__init__.pyi
BvB93: If thus is only exception to the rule, then I feel it might be worthwhile to swap ``generic`` with an union for the time being.

numpy-stubs/__init__.pyi
BvB93: Good catch; an easy fix to implement.

person142:
Thanks @BvB93!

PR #69, opened 2020-04-22
TST: don't run tests in py2 mode

person142:
Since NumPy officially doesn't support Python 2 anymore, there doesn't
seem to be any particular need to test for it. Note that this doesn't
actually do anything to make the stubs not Python 2 friendly as `.pyi`
syntax has always been 2/3 compatible.


rgommers:
LGTM, thanks @person142

PR #70, opened 2020-04-24
TST: also run the code in "pass"

person142:
It seems fair that the code in pass should also run; i.e. that
directory shouldn't contain mypy false positives. Add a test that runs
the modules in that directory and fix the resulting runtime errors.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ setup(
     packages=["numpy-stubs"],
     # PEP 561 requires these
     install_requires=[
-        "numpy>=1.14.0",
+        "numpy>=1.16.0",
         'typing_extensions>=3.7.4; python_version<"3.8"',
     ],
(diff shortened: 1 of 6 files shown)


person142:
Some of the failures are expected in that they are places where the stubs are not get good enough to catch the issue, but some of these might merit issues being opened.

Hm and CI is failing on Python 3.7 only; that's interesting.

person142:
Ok, on 3.7 CI was running NumPy 1.14.2 which is what was causing the failures. I went ahead and bumped the minimum version to 1.16.0; that still goes 3 releases back (soon to be 4), so should be fine I think.

person142:
Fixes for some of the things addressed in this PR have been merged now, so rebased to include them.

rgommers:
Merged, thanks @person142

PR #71, opened 2020-04-24
ENH: Add type annotations for the np.core.fromnumeric module: part 2/4

BvB93:
Add annotations for the second batch of ``np.core.fromnumeric`` functions from https://github.com/numpy/numpy-stubs/pull/59/

New Functions
---------------
* ``argmax()``
* ``argmin()``
* ``searchsorted()``
* ``resize()``
* ``squeeze()``
* ``diagonal()``
* ``trace()``
* ``ravel()``
* ``nonzero()``
* ``shape()``
* ``compress()``

### DIFF
--- a/numpy-stubs/__init__.pyi
+++ b/numpy-stubs/__init__.pyi
@@ -974,7 +974,7 @@ def trace(
     offset: int = ...,
     axis1: int = ...,
     axis2: int = ...,
-    dtype: _DtypeLike = ...,
+    dtype: DtypeLike = ...,
     out: Optional[ndarray] = ...,
 ) -> Union[number, ndarray]: ...
 def ravel(a: ArrayLike, order: _Order = ...) -> ndarray: ...
--- a/numpy-stubs/__init__.pyi
+++ b/numpy-stubs/__init__.pyi
@@ -794,8 +794,8 @@ _ScalarGenericDT = TypeVar(
 _Int = Union[int, integer]
 _Bool = Union[bool, bool_]
 _IntOrBool = Union[_Int, _Bool]
-_ArrayLikeIntNested = Any  # TODO: wait for support for recursive types
-_ArrayLikeBoolNested = Any  # TODO: wait for support for recursive types
+_ArrayLikeIntNested = ArrayLike  # TODO: wait for support for recursive types
+_ArrayLikeBoolNested = ArrayLike  # TODO: wait for support for recursive types
 
 # Integers and booleans can generally be used interchangeably
 _ArrayLikeIntOrBool = Union[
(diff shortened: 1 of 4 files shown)


numpy-stubs/__init__.pyi
```diff
@@ -898,6 +913,23 @@ def partition(
     kind: _PartitionKind = ...,
     order: Union[None, str, Sequence[str]] = ...,
 ) -> ndarray: ...
+@overload
```
BvB93, line 916: The behaviour of ``np.argpartition()`` here is a bit odd.
If an ``np.ndarray`` is passed an ``np.ndarray`` is returned and if a ``np.generic`` is passed a ``np.generic`` is returned, so far so good.

However, if a builtin scalar is passed it also returns a ``np.ndarray``, rather than a ``np.generic``.

```python
>>> import numpy as np

>>> type(np.argpartition(0, 0))
numpy.ndarray

>>> type(np.argpartition(np.int64(0), 0))
numpy.int64
```

person142:
Hey @BvB93, can we move the

> #67 Fixes & Updates

parts into their own PR to keep concerns separate?

BvB93:
> Hey @BvB93, can we move the
> ...
> parts into their own PR to keep concerns separate?

Yeah sure, I've created a pull request for the fixes at https://github.com/numpy/numpy-stubs/pull/74.
In the mean time (_i.e._ until https://github.com/numpy/numpy-stubs/pull/74 is pushed) I'll keep the pull request here as a draft.

BvB93:
With the recent push of https://github.com/numpy/numpy-stubs/pull/74 this pull request is ready for review again.

numpy-stubs/__init__.pyi
```diff
@@ -919,3 +920,69 @@ def argsort(
     kind: Optional[_SortKind] = ...,
     order: Union[None, str, Sequence[str]] = ...,
 ) -> ndarray: ...
+@overload
+def argmax(
+    a: Union[Sequence[ArrayLike], ndarray],
```
person142, line 925: Why not `ArrayLike` here and below?

numpy-stubs/__init__.pyi
```diff
@@ -919,3 +920,69 @@ def argsort(
...
+    axis: int = ...,
+    out: Optional[ndarray] = ...,
+) -> Union[integer, ndarray]: ...
+@overload
+def searchsorted(
+    a: Union[Sequence[ArrayLike], ndarray],
+    v: _Scalar,
+    side: _Side = ...,
+    sorter: Union[None, Sequence[_IntOrBool], ndarray] = ...,  # 1D int array
```
person142, line 952: I don't love the complexity things like this add to each signature versus having some uniform "array like but not a scalar" type that we can use everywhere. This is less detailed for now, but we can reach for this level of detail when making `ndarray`/`ArrayLike` generic; i.e. we could have something like `_NonScalarArrayLike[T]` and keep the code more readable/maintainable. Right now each signature ends up being its own little puzzle.

But, I haven't complained about this on previous PRs, so maybe now is not the time to start objecting.

numpy-stubs/__init__.pyi
```diff
@@ -919,3 +920,69 @@ def argsort(
...
+) -> ndarray: ...
+def resize(a: ArrayLike, new_shape: _ShapeLike) -> ndarray: ...
+@overload
+def squeeze(a: _ScalarGeneric, axis: Optional[_ShapeLike] = ...) -> _ScalarGeneric: ...
+@overload
+def squeeze(a: ArrayLike, axis: Optional[_ShapeLike] = ...) -> ndarray: ...
+def diagonal(
+    a: Union[Sequence[Sequence[ArrayLike]], ndarray],  # >= 2D array
```
person142, line 967: I think we are encountering diminishing returns and increasing code complexity with things like `Sequence[Sequence[ArrayLike]]`.

numpy-stubs/__init__.pyi
BvB93: So the likes of `argmax()`, `argmin()` and `diagonal()` are one of the few cases where, at the very least, a vector or matrix is required. Hence the whole `Sequence[ArrayLike]` and `Sequence[Sequence[ArrayLike]]` syntaxes.

numpy-stubs/__init__.pyi
BvB93: I do like the `_NonScalarArrayLike[T]` suggestion. 

A question though: do you feel it would be better to implement such a change in this pull request? Or shall i create a new pull request afterwards, addressing the issue from both https://github.com/numpy/numpy-stubs/pull/71 and https://github.com/numpy/numpy-stubs/pull/67?

PR #73, opened 2020-04-25
MAINT: add more functions to the blacklist

person142:
Found a few more functions that we shouldn't add types for.

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -16,11 +16,24 @@ BLACKLIST = {
     "numpy": {
         # Stdlib modules in the namespace by accident
         "absolute_import",
+        "division",
+        "print_function",
         "warnings",
-        # Accidentally public
+        # Accidentally public, deprecated, or shouldn't be used
+        "Tester",
         "add_docstring",
         "add_newdoc",
         "add_newdoc_ufunc",
+        "core",
+        "fastCopyAndTranspose",
+        "get_array_wrap",
+        "int_asbuffer",
+        "oldnumeric",
+        "recfromcsv",
+        "recfromtxt",
+        "safe_eval",
+        "set_numeric_ops",
+        "test",
         # Builtins
         "bool",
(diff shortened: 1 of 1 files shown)


runtests.py
```diff
@@ -16,11 +16,24 @@
...
         "add_docstring",
         "add_newdoc",
         "add_newdoc_ufunc",
+        "core",
+        "fastCopyAndTranspose",
+        "get_array_wrap",
+        "int_asbuffer",
+        "oldnumeric",
+        "recfromcsv",
+        "recfromtxt",
```
rgommers, line 33: These two `recfrom` functions are the only ones I'm not sure about, they're not deprecated and are regular data loading functions. Why should they be ignored?

person142: I think I ignored them because they aren't in the refguide and because of https://github.com/numpy/numpy/issues/12385#issuecomment-509505680. Happy to add them back though.

runtests.py
```diff
@@ -41,6 +54,9 @@
         "ppmt",
         "pv",
         "rate",
+        # More standard names should be preferred
+        "alltrue",  # all
+        "sometrue",  # any
```
BvB93, line 59: For ``alltrue`` can't we just use ``alltrue = all`` and call it a day?
Considering it's supposed to be public function (correct me if I'm wrong here) I feel it would not be wise to just skip it.

person142: > Considering it's supposed to be public function

It's not in the refguide, and the general opinion seems to be "should be deprecated/removed" (see https://github.com/numpy/numpy/issues/14584), so I think it's best to exclude it.

runtests.py
person142: Added the two `recfrom` functions back.

person142:
Ok, the two questionable functions have been removed, so merging.

PR #74, opened 2020-04-25
MAINT: Introduced a couple of fixes for the np.core.fromnumeric functions

BvB93:
Follow-up on https://github.com/numpy/numpy-stubs/pull/71#issuecomment-619435429.

https://github.com/numpy/numpy-stubs/pull/67 Fixes & Updates
----------------------------------------------------------------
* Integers and booleans can often be used interchangeably (in fact, ``bool`` is an ``int`` subclass in builtin python). Created``_ArrayLikeIntOrBool`` to reflect this compatibility.
* Rename ``_ScalarGeneric`` to ``_ScalarGenericDT``; add a new ``_ScalarGeneric`` TypeVar which is *only* bound to ``np.generic``.
* ``np.choose()`` can only accept array-like objects consisting of integers (or booleans); reflect this in its annotation.
* ``np.argpartition()`` will return an ``np.integer`` if a scalar is passed; not an ``np.ndarray``.

Further Notes
--------------
* It seems ``np.partition()`` might actually bugged in NumPy ``1.17.3``. 
  it's default value for ``axis`` is set to ``-1`` instead of ``None``. Consequently, it will raise an ``AxisError`` if a scalar is passed *unless* one explicitly passes ``axis=None`` to the function.

```python
>>> import numpy as np

>>> np.partition(0, 0, axis=None)
array([0])

>>> np.partition(0, 0)
  ...
AxisError: axis -1 is out of bounds for array of dimension 0
```


numpy-stubs/__init__.pyi
```diff
@@ -898,6 +913,23 @@ def partition(
     kind: _PartitionKind = ...,
     order: Union[None, str, Sequence[str]] = ...,
 ) -> ndarray: ...
+@overload
```
BvB93, line 916: The behaviour of ``np.argpartition()`` here is a bit odd.
If an ``np.ndarray`` is passed an ``np.ndarray`` is returned and if a ``np.generic`` is passed a ``np.generic`` is returned, so far so good.

However, if a builtin scalar is passed it also returns a ``np.ndarray``, rather than a ``np.generic``.

```python
>>> import numpy as np

>>> type(np.argpartition(0, 0))
numpy.ndarray

>>> type(np.argpartition(np.int64(0), 0))
numpy.int64
```

numpy-stubs/__init__.pyi
```diff
@@ -803,6 +803,7 @@ _Mode = Literal["raise", "wrap", "clip"]
 _Order = Literal["C", "F", "A"]
 _PartitionKind = Literal["introselect"]
 _SortKind = Literal["quicksort", "mergesort", "heapsort", "stable"]
+_Side = Literal["left", "right"]
```
person142, line 806: Looks like this isn't being used?

numpy-stubs/__init__.pyi
```diff
@@ -813,14 +814,27 @@ _ScalarNumpy = Union[generic, dt.datetime, dt.timedelta]
 _ScalarBuiltin = Union[str, bytes, dt.date, dt.timedelta, bool, int, float, complex]
 _Scalar = Union[_ScalarBuiltin, _ScalarNumpy]
 
-_ScalarGeneric = TypeVar(
-    "_ScalarGeneric", bound=Union[dt.datetime, dt.timedelta, generic]
+# Integers and booleans can generally be used interchangeably
+_ScalarInt = TypeVar("_ScalarInt", bound=Union[integer, bool_])
```
person142, line 818: I don't think `_ScalarInt` is the best name here; seems like it should be `_ScalarIntOrBool`. I mostly don't want this to accidentally get cargo-culted to other places where NumPy int/bool diverge; e.g. arithmetic.

numpy-stubs/__init__.pyi
```diff
@@ -813,14 +814,27 @@ _ScalarNumpy = Union[generic, dt.datetime, dt.timedelta]
...
+_ScalarGeneric = TypeVar("_ScalarGeneric", bound=generic)
+_ScalarGenericPlus = TypeVar(
```
person142, line 820: Similarly `_ScalarGenericPlus` is not very descriptive of the type here. Also this definition would allow passing through subclasses of `datetime` and `timedelta`; are we confident that will work correctly?

numpy-stubs/__init__.pyi
```diff
@@ -813,14 +814,27 @@ _ScalarNumpy = Union[generic, dt.datetime, dt.timedelta]
...
+    "_ScalarGenericPlus", bound=Union[dt.datetime, dt.timedelta, generic]
 )
 
 # An array-like object consisting of integers
-_Int = Union[int, integer]
+_Int = Union[int, integer, bool, bool_]
```
person142, line 825: As above about naming.

numpy-stubs/__init__.pyi
```diff
@@ -813,14 +814,27 @@ _ScalarNumpy = Union[generic, dt.datetime, dt.timedelta]
...
+_Bool = Union[bool, bool_]
 _ArrayLikeIntNested = Any  # TODO: wait for support for recursive types
-_ArrayLikeInt = Union[_Int, ndarray, Sequence[_Int], Sequence[_ArrayLikeIntNested]]
+_ArrayLikeBoolNested = Any  # TODO: wait for support for recursive types
+
+# Integers and booleans can generally be used interchangeably
+_ArrayLikeInt = Union[
```
person142, line 831: As above about naming.

numpy-stubs/__init__.pyi
```diff
@@ -862,21 +876,21 @@ def take(
...
+    a: _Int,
     choices: Union[Sequence[_ArrayLike], ndarray],
     out: Optional[ndarray] = ...,
     mode: _Mode = ...,
-) -> _ScalarNumpy: ...
+) -> Union[integer, bool_]: ...
 @overload
 def choose(
-    a: _ArrayLike,
+    a: _ArrayLikeInt,
```
person142, line 893: So while passing bools technically works, given that the the docstring says

"This array must contain integers in [0, n-1], where n is the number of choices"

it seems to me that this is more an accident than the intent. I'm not sure that we should allow this in the types as it seems like not a best practice. @rgommers WDYT about this case? (We could also solicit opinions on the mailing list.)

numpy-stubs/__init__.pyi
person142: Ouch. Sounds like it could potentially be worthy of opening an issue against NumPy.

numpy-stubs/__init__.pyi
BvB93: Good call; this is still a leftover from https://github.com/numpy/numpy-stubs/pull/71.

numpy-stubs/__init__.pyi
BvB93: Accident might not be the best term here.
The general compatibility between ``int`` and ``bool`` stems from the fact the ``bool`` is a subclass of the former. 

```python
>>> import numpy as np

>>> issubclass(int, bool)
True

>>> True == 1 and False == 0
True

>>> np.bool_(True) == 1 and np.bool_(False) == 0
True
```

``Union[int, bool]`` is consequently equivalent to just plain ``int``, but as ``np.bool_`` is not a subclass of ``np.integer`` this does not hold for ``np.bool_``. Even though the latter two classes behave (more or less) identically to ``int`` and ``bool``.

This also means that we _could_ remove ``np.bool_`` from the ``_ArrayLikeInt`` union, but properly removing ``bool`` would require adding a new overload along the lines of 
``choose(a: bool, ...) -> NoReturn``.

numpy-stubs/__init__.pyi
BvB93: Based on https://github.com/numpy/numpy-stubs/pull/74#discussion_r416159856 I'd argue that it is a suitable name. It can still be changed though if you don't agree,

Besides, for arithmetic it would (most likely) be more convenient to use a typevar of ``np.number`` anyway.

numpy-stubs/__init__.pyi
BvB93: Maybe something like ``_ScalarGenericDT`` instead?

About your second concern, some quick testing suggests that subclassing does indeed work:
``` python
>>> import datetime as dt
>>> import numpy as np

>>> class A(dt.timedelta): ...
>>> np.take(A(1), 0).__class__
__main__.A

>>> class B(dt.datetime): ...
>>> np.take(B(2000, 1, 1), 0).__class__
__main__.B
```

numpy-stubs/__init__.pyi
person142: > but as np.bool_ is not a subclass of np.integer this does not hold for np.bool_

And this is why I don't want to include it. Yes, including `int` gets you `bool`, but that's intentionally not true for `numpy.bool_`, so I'm not sure we should make an extra effort to bring it back here. To be clear, when I said "accident" I meant "that `np.bool_` works".

BvB93: Fair enough, I've implemented the name changes in https://github.com/numpy/numpy-stubs/pull/74.

person142:
Thanks @BvB93!

PR #76, opened 2020-04-29
Add more specific integer constructors

### LINKED ISSUE #72: Unsigned int constructors too generous
xref https://github.com/numpy/numpy-stubs/pull/70

The code

```python
import numpy as np

class A:
    def __float__(self):
        return 4.0

np.uint8(A())
```

fails at runtime, but passes type checking. The reason is here:

https://github.com/numpy/numpy-stubs/blob/master/numpy-stubs/__init__.pyi#L402

where the constructor of `number` is:

```python
class number(generic):
    def __init__(self, value: Union[SupportsInt, SupportsFloat] = ...) -> None: ...
```

rajatdiptabiswas:
Fixes #72

- Removed `SupportsFloat` from the classes `int8`, `int16`, `int32`, `int64`, `uint8`, `uint16`, `uint32`, and `uint64`
- Removed a test from `tests/pass/scalars.py`
- Added several tests to `tests/fail/scalars.py`

Kindly let me know if these look alright.


tests/pass/scalars.py
```diff
@@ -20,7 +20,6 @@ def __float__(self):
 np.int16(3.4)
 np.int32(4)
 np.int64(-1)
-np.uint8(A())
```
person142, line 23: Instead of removing, I would follow what is done here:

https://github.com/numpy/numpy-stubs/pull/70/files#diff-dd377d4c09025262e6f7008bcfe583dfR10

i.e.

```python
class B:
    def __int__(self):
        return 4

...

np.uint8(B())
```

(That way we keep a case checking that a class with `__int__` passes.)

rajatdiptabiswas:
Made the changes you mentioned @person142. 
Take a look at it and let me know if everything looks fine now.

Thanks.

person142:
Merged, thanks @rajatdiptabiswas.

PR #78, opened 2020-05-06
MAINT: Added ndarray-subclass support to the ndarray methods 

BvB93:
* Added ``ndarray``-subclass support to the ``ndarray`` methods 
* ``flatiter`` is now a generic with respect to ``_ArrayOrScalarCommon``.
* ``flatiter`` is now a subclass of ``Iterator[generic]``.
* Changed ``ndarray.real`` and ``ndarray.imag`` into get- and set-supporting properties.
* Removed a single redundant overload from ``ndarray.view`` after complaints from mypy.
* Fixed a typo in ``ndarray.view``: one of the overloads had the ``type`` argument marked as ``dtype``.


numpy-stubs/__init__.pyi
```diff
@@ -337,47 +345,51 @@ class ndarray(_ArrayOrScalarCommon, Iterable, Sized, Container):
...
+    def view(self, type: Type[_NdArraySubClass]) -> _NdArraySubClass: ...
     @overload
-    def view(self, dtype: _DtypeLike = ...) -> ndarray: ...
+    def view(self: _ArraySelf, dtype: _DtypeLike = ...) -> _ArraySelf: ...
     @overload
     def view(
         self, dtype: _DtypeLike, type: Type[_NdArraySubClass]
     ) -> _NdArraySubClass: ...
-    @overload
-    def view(self, *, type: Type[_NdArraySubClass]) -> _NdArraySubClass: ...
```
BvB93, line 358: This particular overload is redundant.
The signature described here expects ``type`` as a keyword-only argument, yet the very first overload already expects ``type`` as a keyword-or-positional argument.

numpy-stubs/__init__.pyi
```diff
@@ -337,47 +345,51 @@ class ndarray(_ArrayOrScalarCommon, Iterable, Sized, Container):
...
         subok: bool = ...,
         copy: bool = ...,
-    ) -> ndarray: ...
-    def byteswap(self, inplace: bool = ...) -> ndarray: ...
-    def copy(self, order: str = ...) -> ndarray: ...
+    ) -> _ArraySelf: ...
+    def byteswap(self: _ArraySelf, inplace: bool = ...) -> _ArraySelf: ...
+    def copy(self: _ArraySelf, order: str = ...) -> _ArraySelf: ...
     @overload
-    def view(self, dtype: Type[_NdArraySubClass]) -> _NdArraySubClass: ...
```
BvB93, line 350: A type; this should be the ``type`` argument, not ``dtype``.

numpy-stubs/__init__.pyi
```diff
@@ -294,17 +296,23 @@ class _ArrayOrScalarCommon(
 _BufferType = Union[ndarray, bytes, bytearray, memoryview]
 
 class ndarray(_ArrayOrScalarCommon, Iterable, Sized, Container):
-    real: ndarray
-    imag: ndarray
+    @property
```
BvB93, line 301: The documentation for ``ndarray.real`` and ``ndarray.imag`` ([ref](https://numpy.org/doc/stable/reference/generated/numpy.real.html)) suggests that both getting and setting should be fine.

numpy-stubs/__init__.pyi
```diff
@@ -179,16 +181,16 @@ class _flagsobj:
     def __getitem__(self, key: str) -> bool: ...
     def __setitem__(self, key: str, value: bool) -> None: ...
 
-class flatiter:
+class flatiter(Generic[_ArraySelf], Iterator[generic]):
```
BvB93, line 186: Once ``ndarray`` is a proper Generic we can get more specific about the iterators elements; for now it's set to ``np.generic``.

numpy-stubs/__init__.pyi
```diff
@@ -130,16 +132,18 @@ class _flagsobj:
     def __getitem__(self, key: str) -> bool: ...
     def __setitem__(self, key: str, value: bool) -> None: ...
 
-class flatiter:
+_FlatIterSelf = TypeVar("_FlatIterSelf", bound=flatiter)
+
+class flatiter(Generic[_ArraySelf], Iterator[generic]):
```
person142, line 137: I don't think you need the inheritance from `Iterator` here; the `__iter__` and `__next__` defined below should cover it.

numpy-stubs/__init__.pyi
```diff
@@ -246,17 +250,23 @@ class _ArrayOrScalarCommon(
 _BufferType = Union[ndarray, bytes, bytearray, memoryview]
 
 class ndarray(_ArrayOrScalarCommon, Iterable, Sized, Container):
-    real: ndarray
-    imag: ndarray
+    @property
+    def real(self: _ArraySelf) -> _ArraySelf: ...
+    @property
```
person142, line 255: Shouldn't this be `@real.setter`?

numpy-stubs/__init__.pyi
```diff
@@ -246,17 +250,23 @@ class _ArrayOrScalarCommon(
...
+    def real(self, value: ArrayLike) -> None: ...
+    @property
+    def imag(self: _ArraySelf) -> _ArraySelf: ...
+    @property
```
person142, line 259: Same comment as above.

numpy-stubs/__init__.pyi
BvB93: Oh wow. Yes, yes it absolutely should be.

numpy-stubs/__init__.pyi
BvB93: I've removed it as of https://github.com/numpy/numpy-stubs/pull/78/commits/b5341a592d414ad26486f2ecec2a6a3e030a5e44.

person142:
Thanks @BvB93!

Issue #65, opened 2020-04-21
Stubs in `core/numerictypes.pyi` are not picked up


person142:
This is the same issue as https://github.com/numpy/numpy-stubs/pull/50, but for `core/numerictypes.pyi`. The stubs are not imported into `__init__.pyi`, so they are not picked up. A quick test of that:

```python
import numpy as np

reveal_type(np.maximum_sctype(np.dtype(np.float64)))
```

gives `Revealed type is 'Any'` on master. We should move those stubs into the top-level `__init__.pyi` and add tests.

PR #83, opened 2020-05-24
MAINT: pick up types from `core/numerictypes.pyi`

person142:
Closes https://github.com/numpy/numpy-stubs/issues/65.

The types currently aren't picked up. Fix that by moving them into
`__init__.pyi` along with the rest of the types. While we are at it,
polish the signatures a bit and add tests.


person142:
Unless someone objects, I will merge this in a couple of days. (I don't want to port this unconnected file to NumPy.)

